PR #142, opened 2026-06-08
fix(cart): serialize cart cache as JSON so Redis writes stop falling back to MongoDB (#141)

### LINKED ISSUE #141: cart: Redis cart cache write always fails (datetime not JSON serializable) → 100% MongoDB fallback
## Summary
The cart cache is designed as a two-tier store: **Redis** (Dapr state store `cartstore`) as primary, with the **MongoDB `cache_cart` collection** as a circuit-breaker fallback. In practice the Redis write fails on **every** request because the cart document is serialized with `datetime` objects that aiohttp's default JSON encoder cannot handle. As a result, **every cart falls back to MongoDB** and Redis caching provides zero benefit.

## Evidence (observed on a running stack)
- Redis db2 (`cartstore`) is **empty** (`DBSIZE = 0`) — no cart is ever stored in Redis.
- `services-cart` logs show, on **every** cache write:
  ```
  [WARNING] app.models.repositories.cart_repository: Failed to cache cart: Object of type datetime is not JSON serializable
  ```
- All in-progress carts (`transaction_no: -1`, status `EnteringItem`/`Paying`/...) accumulate in `db_cart_<tenant>.cache_cart`.

## Root cause
`services/cart/app/models/repositories/cart_repository.py` → `__cache_cart_async`:

```python
cart_data = cart.model_dump()                      # datetimes stay as datetime objects
state_post_data = [{"key": cart.cart_id, "value": cart_data}]
async with session.post(self.base_url_cartstore, json=state_post_data) as response:  # aiohttp default json.dumps
```

`cart.model_dump()` (no `mode="json"`) returns Python `datetime` values for `created_at`/`updated_at` and for the embedded master snapshot (`masters.promotions[].start_datetime` / `end_datetime`). aiohttp serializes `json=` with the stdlib encoder, which raises on `datetime`. The exception is caught by `cache_cart_async` and falls back to `__save_cart_to_db_async` (MongoDB), which succeeds because Motor handles BSON dates natively.

Because `created_at`/`updated_at` are always datetimes, this path has likely been broken for a long time; it goes unnoticed because the MongoDB fallback keeps cart operations functionally working (the circuit breaker just oscillates open/half-open).

## Impact
- Redis cart cache is effectively disabled — all cart reads/writes hit MongoDB (performance regression vs. the intended design).
- `cache_cart` grows unbounded with in-progress carts and is never cleaned up (see secondary issues below).

## Proposed fix (primary)
Serialize datetimes for the Dapr/Redis payload:

```python
cart_data = cart.model_dump(mode="json")   # datetime -> ISO strings (JSON-serializable)
```

Round-trip is safe: `__get_cached_cart_async` does `CartDocument(**cart_data)`, and pydantic parses ISO strings back to datetime.

## Secondary cleanup gaps (orphaned MongoDB fallback data)
Even once Redis works, fallback copies can leak:
- `delete_cart_async` deletes from **Redis only** when the circuit is closed; it does not remove the MongoDB `cache_cart` copy → orphaned docs.
- `cache_cart` has no TTL index (only `_id` and `cart_id`), unlike Redis `cartstore` (TTL 10h), so fallback data never expires.

Consider: (a) delete the MongoDB fallback copy on `delete_cart_async`, and/or (b) add a TTL index on `cache_cart` (e.g. `updated_at` with `expireAfterSeconds`).

## Notes
- Unrelated to PR #140 / #139 (receipt print data); filing separately for independent handling.

kugel-masa:
## Summary
Fixes #141. The cart two-tier cache (Redis primary via Dapr `cartstore`, MongoDB `cache_cart` as circuit-breaker fallback) was writing the cart with `cart.model_dump()`, leaving `datetime` objects in the payload (`created_at`/`updated_at` and the embedded promotion master snapshot). aiohttp's stdlib json encoder then raised `Object of type datetime is not JSON serializable` on **every** write, so every cart silently fell back to MongoDB — Redis caching provided zero benefit and the circuit breaker oscillated open/half-open.

### Primary fix
- `cart_repository.__cache_cart_async`: use `model_dump(mode="json")` so datetimes become ISO strings. Round-trip is safe — `__get_cached_cart_async` rebuilds via `CartDocument(**data)` and pydantic parses ISO strings back to datetime.

### Secondary cleanup (orphaned MongoDB fallback data)
- `delete_cart_async` now best-effort removes the MongoDB fallback copy after a successful Redis delete (non-raising), so `cache_cart` no longer leaks orphaned docs.
- Added a `created_at` TTL index on `cache_cart`, aligned with the Redis `cartstore` 10h TTL (`CACHE_CART_TTL_SECONDS=36000`). `created_at` is used because `updated_at` can be `None` on first insert and would leave such docs unexpired. The index is only added when the collection is first created (existing tenants are not back-filled).
- `commons`: `create_indexes_command` / `create_collection_with_indexes_async` gain an optional `expire_after_seconds` (`expireAfterSeconds`) parameter — backward compatible. Version bumped to 0.1.53 and propagated to all services.

## Testing
All 7 services green across every tier:
- **unit**: cart 515, commons 359 (incl. new TTL command-builder tests)
- **integration**: 7/7 services pass (real MongoDB)
- **e2e**: 8/8 (all services + cross-service) pass on the rebuilt full stack

### Live verification (Issue's core symptom)
On the rebuilt stack during e2e, cart container logs/DB:
| Metric | Before | After |
|---|---|---|
| `Cart cached` (Redis write success) | 0 | **206** |
| `Failed to cache cart` (datetime error → fallback) | every request | **0** |
| `Circuit is open` | frequent | **0** |
| `cache_cart` orphaned docs | unbounded | **0** |
- TTL index confirmed in live MongoDB: `cache_cart_index_created_at -> expireAfterSeconds: 36000`


kugel-masa (commented):
## Code Review Summary (manual, full diff excl. Pipfile.lock)

### Findings

#### 🔴 Critical — none
#### 🟡 Warning — none

#### 🔵 Suggestion (Nice to Have)
- **[Performance]** `cart_repository.py` — `__delete_db_fallback_copy_async` runs **unconditionally** on every cart delete. `delete_async` (commons) does a `find_one` first and returns `False` on no match, so after the primary fix (where a fallback copy almost never exists) every cart deletion incurs one extra MongoDB `find_one` round-trip. Cheap (cart_id is uniquely indexed) and only at transaction-completion frequency, so acceptable — but a known tradeoff. Could be made conditional by tracking whether a fallback ever occurred (at the cost of extra state; current simplicity is defensible).
- **[Observability]** Same path — in the common no-fallback case, `delete_async` emits an `INFO` "Document not found..." on every cart delete. Not WARNING-level spam, but minor INFO noise.
- **[Ops/Migration]** `create_collection_with_indexes_async` only adds indexes when the collection is first created (`if created`), so **existing tenant DBs will not get the new `created_at` TTL index** (an agreed limitation). Recommend a one-off index-creation script/runbook note for existing environments.
- **[Pre-existing/Out of scope]** `cart_repository.py` still has unrelated pre-existing lint (unused `aiohttp` import, duplicate `settings` import). Not introduced here; flagging as a future cleanup candidate.

#### ✅ Good Practices
- `expire_after_seconds is not None` guard correctly allows `expireAfterSeconds=0` (immediate expiry); covered by an explicit test.
- commons API change is fully backward compatible (optional param only).
- Cleanup is non-raising, preserving the "cache delete succeeded" contract; circuit-open path does not double-delete (covered by `test_delete_does_not_clean_db_when_circuit_open`).
- Regression tests pin the core behavior (JSON-serializability + round-trip + delete cleanup).
- TTL keyed on `created_at` rather than `updated_at` (which can be `None` on first insert and would leave docs unexpired) — rationale documented in code.
- Verified live: 206 successful Redis cache writes, 0 datetime errors, 0 orphaned `cache_cart` docs, TTL index confirmed in MongoDB.

### Verdict: ✅ Approve
No blockers — findings are future optimizations / ops notes only. Fix is correct with tests and live verification.

Issue #139, opened 2026-06-07
receipt-print-schema: device-agnostic JSON for receipt print data (remove XML, expand expressiveness)


kugel-masa:
## Summary

Redesign receipt print data into a **device-agnostic JSON format** that does not depend on OPOS or a specific printer model. Remove the current XML (`receipt_text`) and fully migrate to JSON (`print_document`). At the same time, expand expressiveness (alignment, character styling such as double-width, logo, image, barcode, QR, line feed, paper cut).

> The scope of this issue is **the format of the print data produced by the application (backend) and the logic that generates it**. Converting/rendering the print data into actual device commands (a device-gateway-equivalent) is out of scope for this repository; final rendering is the responsibility of the print-data consumer (frontend / device).

---

### 1. XML → JSON is a hard cutover (no parallel operation)
- Remove the current `receipt_text` (XML) and switch entirely to JSON (`print_document`). Keep `journal_text` (electronic journal, plain text).
- Rationale: avoid storing the same content twice in both XML and JSON, **reducing data volume** (storage, pub/sub, data sync).
- Do not migrate existing historical data (tranlog / journal_document that hold XML); allow mixed formats on the read side.

### 2. R / J / RJ channels (electronic routing)
- As the backend's electronic routing axis: R/RJ → receipt content (print-data JSON), J/RJ → electronic journal (`journal_text`).
- Physical journal printing is out of scope (electronic journal only).

### 3. Schema (emit columns as semantic information)
- A line element has `text` (whole-line) and `columns` (multi-column line).
- `columns` has three slots: **left / mid / right**. `mid` specifies its start position via a fixed offset from the left, `startCol`.
- Character styling (double-width, emphasis, underline, reverse, font) can be specified at **`text` = line level / `columns` = column level** (e.g., make only the amount column double-width). Sub-column span granularity is out of scope.
- The backend emits **semantic columns** (each column's value, position, and styling) and does not produce pre-formatted text with baked-in spaces. Final column alignment (space calculation, placement, wrapping) and rendering are the **consumer's responsibility**.
- Carry a design-reference width as `metadata.charsPerLine`; the consumer uses it as the basis to adapt to the actual printer width.

### 4. Character encoding
- Keep the pipeline in **UTF-8**. Conversion to the target device's character set (e.g., Shift-JIS) is the responsibility of the consumer / device.
- Characters that cannot be mapped on the target side should be replaceable with substitute characters by the consumer so printing can continue.

[...]

PR #145, opened 2026-06-08
feat(receipt): add explicit width to mid column so align is well-defined (#143)

### LINKED ISSUE #143: receipt-print-schema: add explicit width to mid column so align is well-defined (follow-up to #139)
## Summary

Give the `mid` column an explicit **width** (in addition to `startCol`) so that per-column `align` is well-defined for it. Follow-up to #139 (device-agnostic print-data JSON).

## Problem

In the current `columns` schema (#139), a `Column` has `slot` (`left`/`mid`/`right`), `value`, `startCol` (for `mid`), `align`, and `style` — but **no explicit width**. Cell boundaries are only *implicit*:

- `left` starts at col 0 and runs to the next anchor,
- `right` is anchored to the right edge (cell = content width),
- `mid` starts at `startCol` but has **no defined end**, so the region that `align` should operate within is undefined.

As a result, `align` on a `mid` column cannot be realized: without a width there is no cell to align inside.

## Proposed change

Add a **`width`** field (half-width columns) used by the `mid` slot:

- `mid` carries `startCol` + `width` → its cell is `[startCol, startCol + width)`.
- `align` applies **within that cell**:
  - `align="right"` → right-justify the value inside the cell; the left padding from `startCol` up to the value is filled with spaces.
  - `align="left"` (default) → value at `startCol`.
  - `align="center"` → centered within the cell.
- `left` (col-0 start, cell to next anchor) and `right` (right-edge anchor, cell = content) keep their current anchor-based behavior; only `mid` gets an explicit width.

### Example (charsPerLine=48)

```
お茶 500ml        @108 x1   108外
                  └ mid: startCol=14, width=10, align=right
                    (value right-justified inside [14, 24), left side padded with spaces)
```

## Scope

- `print_document_model.py`: add `Column.width: Optional[int]` (serialized as semantic info; no baked spaces in the JSON).
- `abstract_receipt_data.py`: in the plain-text journal rendering, honor `mid` `width` + `align` (right-justify with left-space padding, etc.).
- `contracts/print-document.schema.md`: document `mid` = `startCol` + `width`, and that `align` operates within `[startCol, startCol+width)` (`right` → left padded with spaces). Clarify `left`/`right` cell derivation.
- `examples/` + tests: show a `mid` column with `width` + `align=right`.

## Non-regression

Production receipt generators only use `left` + `right` (from `line_split`); none use `mid`. So `journal_text` for production receipts stays byte-equivalent — this change only affects receipts that opt into `mid` columns (examples/tests today).

## Notes

- The final visual layout/rendering remains the consumer's responsibility; this only makes the `mid` cell (and thus its `align`) well-defined in the schema.
- Out of scope: arbitrary fixed-width N-column tables (would need per-column widths on all slots / a dedicated table element).

kugel-masa:
## Summary

Follow-up to #139. The `columns` schema gave a `mid` column a `startCol` but **no width**, so the region that `align` should operate within was undefined — `align` on a `mid` column could not be realized. This adds an explicit **`width`** field so a `mid` cell is `[startCol, startCol + width)` and `align` is well-defined inside it.

Closes #143.

## Changes

- **`print_document_model.py`**: add `Column.width: Optional[int]` (`ge=0`, serialized camelCase as semantic info; no baked spaces in the JSON).
- **`abstract_receipt_data.py`**: in plain-text journal rendering, honor `mid` `width` + `align` via a new `_render_mid_cell` helper:
  - `align="right"` → right-justify inside the cell (left padded with spaces)
  - `align="center"` → centered (odd remainder favors the right)
  - `align="left"` (default) → value at `startCol`, right padded
  - value wider than the cell overflows right with no extra padding
  - padding is computed from **display width (`wcwidth`)**, so full-width/Japanese values land at the correct columns
  - `mid` without `width` keeps legacy behavior (value at `startCol`, `align` ignored)
- **`contracts/print-document.schema.md`**: document the `width` field, `left`/`mid`/`right` cell derivation, and `align`-within-cell semantics.
- **examples**: `normal-sale.json` and `builder-example.py` show a `mid` column with `width: 10, align: "right"` (the issue's `お茶 500ml` example). Builder output is byte-identical to `normal-sale.json`.
- **tests** (`test_receipt_columns_mid_width.py`): cell alignment (right/left/center/overflow), display-column positioning, full-width (Japanese) handling, legacy no-width path, left+right pair unchanged, and JSON serialization.

## Non-regression

Production receipt generators only use `left` + `right` (from `line_split`); none use `mid`. So `journal_text` for production receipts stays byte-equivalent — this change only affects receipts that opt into `mid` columns (examples/tests today).

## Verification

- commons unit suite: **373 passed** (including 10 new mid-width tests).
- `builder-example.py` output `diff` against `normal-sale.json`: **MATCH**.
- Full-width handling verified: rendered `mid` cells are exactly `width` display columns wide and land at the expected `startCol`.

### DIFF
--- a/services/commons/src/kugel_common/receipt/abstract_receipt_data.py
+++ b/services/commons/src/kugel_common/receipt/abstract_receipt_data.py
@@ -171,12 +171,37 @@ class AbstractReceiptData(ABC, Generic[T]):
             if start > cursor:
                 line += TextHelper.space(start - cursor)
                 cursor = start
-            line += column.value
-            cursor += wcwidth.wcswidth(column.value)
+            # A `mid` column with an explicit width defines a cell
+            # [startCol, startCol + width) that `align` operates within;
+            # otherwise the value is placed as-is at its start (legacy mid/left).
+            if column.slot == "mid" and column.width is not None:
+                cell = self._render_mid_cell(column.value, column.width, column.align)
+                line += cell
+                cursor += wcwidth.wcswidth(cell)
+            else:
+                line += column.value
+                cursor += wcwidth.wcswidth(column.value)
         if right_value is not None:
             pad = width - cursor - wcwidth.wcswidth(right_value)
             line += TextHelper.space(max(1, pad)) + right_value
         return line
+
+    @staticmethod
+    def _render_mid_cell(value: str, cell_width: int, align: Optional[str]) -> str:
+        """Render a `mid` column value inside its [startCol, startCol+width) cell.
+
+        ``align`` is honored within the cell: ``right`` left-pads with spaces,
+        ``center`` centers, ``left`` (default) right-pads. A value wider than the
+        cell overflows to the right with no padding on the overflow side."""
+        pad_total = max(0, cell_width - wcwidth.wcswidth(value))
+        if align == "right":
+            left_pad, right_pad = pad_total, 0
+        elif align == "center":
+            left_pad = pad_total // 2
+            right_pad = pad_total - left_pad
+        else:  # left / default
+            left_pad, right_pad = 0, pad_total
+        return TextHelper.space(left_pad) + value + TextHelper.space(right_pad)
 
     #
     # date/format helpers
--- a/services/commons/src/kugel_common/receipt/print_document_model.py
+++ b/services/commons/src/kugel_common/receipt/print_document_model.py
@@ -80,6 +80,11 @@ class Column(BaseModel):
     # Required only for the `mid` slot: fixed offset from the left (half-width
     # columns, 0-based, based on metadata.charsPerLine).
     start_col: Optional[int] = Field(None, alias="startCol")
+    # `mid` only: explicit cell width (half-width columns). With it the `mid`
+    # cell is [startCol, startCol + width) and `align` operates within that cell
+    # (right -> left-padded with spaces, center -> centered). `left`/`right`
+    # derive their cell from anchors and ignore `width`.
+    width: Optional[int] = Field(None, ge=0)
     # Default per slot: left/mid -> left, right -> right (resolved by consumer).
     align: Optional[Literal["left", "center", "right"]] = None
     style: Optional[Style] = None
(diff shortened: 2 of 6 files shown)


kugel-masa:
## Code Review & Resolutions

A self-review was performed against the diff, surrounding code, and project conventions. Verdict: **Approve**. Findings and how each was handled below.

### Strengths
- Padding is display-width based (`wcwidth.wcswidth`), so full-width/Japanese values keep the cell exactly `width` columns wide.
- `center` rounding (`pad//2` left, remainder right) matches the existing `TextHelper.fixed_center` convention — text-center and mid-cell-center stay consistent.
- Backward compatible: `mid` without `width` keeps legacy behavior; the production `left+right` fast path is untouched (journal stays byte-equivalent).

### Findings & resolutions

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | minor | No test for a wide `mid` cell colliding with the `right` column. | **Fixed** — added `test_mid_width_colliding_with_right_keeps_min_one_space` (commit `6ebca22`), locking in the `max(1, pad)` min-one-space behavior (no truncation, no crash). |
| 2 | nit | Model allows `width` on `left`/`right` (ignored at render time); no `mid`-only validation. | **By design / wontfix** — consistent with the existing lenient `startCol` (also not enforced mid-only). Documented in schema.md that `left`/`right` ignore `width`. |
| 3 | minor | Journal renderer does not truncate on column collision (schema mentions truncation). | **By design** — truncation is the consumer's responsibility; the general column path never truncated pre-#143. Behavior preserved. |
| 4 | nit | Trailing spaces when `mid` is `align=left/center` with no following `right`. | **No-op** — pre-existing in journal_text (`_render_text_element` pads to full width); not a regression. |
| 5 | info | `wcswidth` returns `-1` for control chars (no `None`/`-1` guard in `_render_mid_cell`). | **No-op** — consistent with existing `_render_columns_general`, which uses `wcswidth` directly; receipt values contain no control chars in practice. |

### Verification
- commons unit suite: **376 passed** (including 11 mid-width tests).
- `builder-example.py` output `diff` vs `normal-sale.json`: **MATCH**.

Issue #153, opened 2026-06-12
cart: unit-test patch leak in test_cart_service.py breaks integration tests in the same pytest process


kugel-masa:
Found while implementing #148 (not introduced by it).

## Problem

\`services/cart/tests/unit/test_cart_service.py\`'s \`_build_service()\` helper installs a module-level patch and never stops it:

```python
# Install persistent patch on calc_subtotal_logic (stays active for test execution)
patcher = patch("app.services.cart_service.calc_subtotal_logic")
mock_subtotal_logic = patcher.start()
mock_subtotal_logic.calc_subtotal_async = AsyncMock(side_effect=lambda cart, repo: cart)
```

\`patcher.stop()\` is never called, so after the unit tier runs, \`app.services.cart_service.calc_subtotal_logic\` remains a \`MagicMock\` whose \`calc_subtotal_async\` returns the cart unchanged — for the rest of the pytest process.

## Symptom

Running unit and integration tiers **in one process** fails integration tests that exercise payments/billing, e.g.:

```
pipenv run pytest tests/unit tests/integration
# FAILED tests/integration/test_cart_operations.py::test_bill_full_payment
#   406 {"user_error":{"code":"403001"...}} "The balance is equal to 0 ...
#   payments: [{'payment_code': '01', 'amount': 100}], add_payment: {'payment_code': '01', 'amount': 100}"
```

With the subtotal/balance recalculation mocked out, the second payment application sees a stale balance and the cashier flow double-applies. 7 tests fail in the combined run; each tier passes in isolation.

## Why it goes unnoticed

\`scripts/run_unit_tests.sh\` and \`scripts/run_integration_tests.sh\` run the tiers in separate pytest processes, so the leak never crosses over in CI. It only bites ad-hoc combined invocations (\`pytest tests/unit tests/integration\`).

## Fix direction

Stop the patcher deterministically — any of:

- Convert \`_build_service\` into a fixture (or pair it with one) that calls \`patcher.stop()\` on teardown.
- Use \`with patch(...)\` scoping like the \`CartStrategyManager\` / \`PromotionMasterWebRepository\` patches in the same helper already do.
- As a backstop, add \`unittest.mock.patch.stopall()\` in a \`tests/unit/conftest.py\` autouse teardown.

## Repro

```
cd services/cart
pipenv run pytest tests/unit/test_cart_service.py "tests/integration/test_cart_operations.py::test_bill_full_payment"
[...]

Issue #152, opened 2026-06-12
cart: idempotent transaction finalization keyed on cart_id (replay double-count guard)


kugel-masa:
Split out of #148 (spec analysis finding; see \`specs/148-cart-snapshot-restore/spec.md\` Clarifications 2026-06-12). Part of the #146 client-carried cart design.

## Background

The transaction log (\`BaseTransaction\` / tranlog) does **not** carry \`cart_id\`. Uniqueness is enforced on \`(tenant_id, store_code, terminal_no, transaction_no)\`, and \`transaction_no\` is newly issued on every bill. \`cart_id\` exists only as the cache key of in-flight carts (\`cache_cart\`), which is gone once the transaction is finalized.

With the #148 restore API, this leaves a structural double-count path:

1. POS holds a **pre-terminal** snapshot (status=paying) of a cart.
2. The transaction is finalized (bill) — cart leaves the cache.
3. The old snapshot is restored (valid signature, cart no longer exists server-side → restore succeeds; the terminal-state rejection \`401506\` does not apply because the snapshot itself is not terminal).
4. Bill again → a **new** \`transaction_no\` is issued → the same sale is counted twice.

\#148's posture is replay **accept + detect**: it ships terminal-state idempotent rejection and a full restore audit trail (\`log_cart_restore\`), so the double-count is detectable but not prevented. Prevention is this issue.

## What

1. **Carry \`cart_id\` into the transaction log** (tranlog document + persisted collection).
2. **Idempotent finalization**: at bill time, look up an existing finalized transaction with the same \`cart_id\` (tenant/store scope). If found, respond idempotently (return the existing transaction outcome / dedicated error code) instead of issuing a new \`transaction_no\` and writing a second tranlog.
3. **Index**: add an index on \`(tenant_id, store_code, cart_id)\` to make the check cheap; decide unique vs non-unique considering void/return transactions and the race between concurrent bills (the downstream design treats \`cart_id\` as the transaction identity key).
4. **Tests**: restore a pre-bill snapshot of an already-finalized transaction → second bill is idempotent, tranlog count unchanged. This completes **SC-004 of #148** (exactly one recorded transaction per cart_id).

## Notes

- Historical tranlog documents have no \`cart_id\`; readers must tolerate its absence (no migration).
- Numbering context carried in snapshots (the downstream solution for stateless finalization) is **out of scope** here — tracked under #146 phase 2.

## Dependencies

[...]

Issue #147, opened 2026-06-11
all services: enable gzip response compression (client-carried cart phase 0)


kugel-masa:
Phase 0 of #146.

## What

Add response gzip compression to all 7 services. Currently no service has compression middleware (only CORS and request logging).

```python
from fastapi.middleware.gzip import GZipMiddleware

app.add_middleware(GZipMiddleware, minimum_size=1024, compresslevel=6)
```

- Compresses only when the client sends `Accept-Encoding: gzip`; harmless to clients that don't.
- Decompression is automatic in essentially every HTTP client, so no frontend change is required.
- `compresslevel=6` is the speed/ratio sweet spot (9 burns CPU for a few extra percent).

Prefer wiring it through a shared setup in `kugel_common` (same placement pattern as `log_requests`) if it stays a one-liner per service either way.

## Why now

- Zero behavior change, immediately useful on its own (responses already include `receipt_text` and full cart bodies).
- Prerequisite for measuring the compressed size of cart snapshots before Phase 1/2 decisions (#146).

## Scope

- account, terminal, master-data, cart, report, journal, stock: add middleware
- Note: request-body decompression (`Content-Encoding: gzip` from the POS) is **out of scope** — that is Phase 2 and needs a custom ASGI middleware in kugel_common with a decompressed-size guard.

Issue #146, opened 2026-06-11
cart: client-carried cart_document — make the cart backend stateless (design umbrella)


kugel-masa:
## Background

Load testing in a downstream deployment surfaced a weakness that is not downstream-specific but inherent to the upstream fallback design: when a cart read against the Dapr state store times out, the code falls back to the MongoDB `cache_cart` collection — which is empty during normal operation — so an existing in-flight cart turns into a 404 business error.

## Weaknesses of the current design

The current design philosophy is: "if Redis breaks, new transactions can continue via MongoDB fallback; in-flight transactions are lost and must be re-entered." Structurally (`services/cart/app/models/repositories/cart_repository.py`):

- During normal operation the cart is written only to the Dapr state store (Redis); the MongoDB fallback collection `cache_cart` stays empty.
- A failed read falls back to the DB immediately, with no retry (lines 190, 202) → reads the empty collection → 404. **A transient timeout turns an existing cart into a business error.**
- The circuit breaker opens after 3 consecutive failures (`_failure_threshold = 3`, line 46). Below the threshold (flapping), failed writes go to the DB while reads still prefer Redis, so **a stale Redis copy can win over a newer MongoDB copy — the transaction silently rolls back**.
- If a Redis failure involves data loss, in-flight carts have no replica anywhere; fallback cannot save them by design.

So the fallback only protects "accepting new transactions", never "continuing in-flight transactions".

## Direction: client carries the cart_document (stateless backend)

Instead of hardening the fallback (e.g. permanent dual writes), solve the category of problems at the root: **the frontend carries the cart_document in every request/response round trip.**

- The backend becomes a logic machine that holds no cart state.
- A POS can continue a transaction mid-flight against any edge/cloud backend it connects to (good fit for multi-edge stores and backends with no POS-affinity).
- The Redis/MongoDB cache problems (timeout→404, split-brain, meaningless fallback) disappear as a category. Server-side cache becomes an optional performance optimization (or is removed), no longer authoritative.
- A downstream failover feature already adopted the same idea limited to connection switchover (every response carries a signed cart snapshot; the new backend restores from it). This issue generalizes it to every request.

## Verified facts (against current upstream code)

**`cart.masters` can be carried as-is — a version-reference scheme is rejected.**

- `cart.masters.items` is not a catalog replica; it is a **session snapshot of only the items scanned in this cart** (`item_master_web_repository.py`, `cart_service.py:792`).
[...]

PR #154, opened 2026-06-12
feat(cart): signed cart snapshot in responses + restore API (client-carried cart phase 1, #148)

### LINKED ISSUE #148: cart: attach signed cart snapshot to responses + restore API (client-carried cart phase 1)
Phase 1 of #146. Port of a failover design proven in a downstream deployment (every response carries a signed cart snapshot; a new backend restores from it).

## What

1. Every cart-mutating response additionally carries a **signed snapshot of the full cart_document** (including `cart.masters` — see #146 for why masters are carried as-is, not by version reference).
2. A **restore API** accepts a signed snapshot and re-materializes the cart on a backend that has never seen it, after verifying the signature.

After this phase:

- A POS can switch to another edge/cloud backend mid-transaction and continue (present the last snapshot to the new backend).
- "Redis data loss ⇒ in-flight transactions must be re-entered" is gone — the frontend always holds a restorable copy.

The server-side cache (Redis/MongoDB fallback) stays authoritative in this phase; full statelessness is Phase 2.

## Design decisions to make (spec before implementation)

- **Signature**: HMAC over the snapshot; key management/rotation across services; what exactly is signed (canonical serialization).
- **Snapshot schema**: full `CartDocument` vs. a dedicated transport schema; versioning for forward compatibility.
- **Restore API**: endpoint shape, authz (terminal API key + tenant scoping), behavior when a cart with the same cart_id already exists (winner rule — the downstream design uses cart_id as the transaction identity key).
- **Replay**: an older signed snapshot is valid by construction; posture is accept + detect (idempotent finalization on cart_id + audit trail), not prevent.
- **Size/latency budget**: validate the desk estimate (~5–15 KB gzipped for a 40-item cart) with measurements; depends on #147 for compressed sizes. Production data from the downstream deployment feeds this.

## Plan

Spec first via speckit (`specs/`), then implement. Implementation should not start before the signature and restore-conflict semantics above are pinned down.

kugel-masa:
Closes #148. Phase 1 of the #146 client-carried cart design: every cart-mutating response carries a signed snapshot of the full cart document, and a new restore API re-materializes the cart on a backend that has never seen it.

## What

**Spec kit first** (`specs/148-cart-snapshot-restore/`): spec → plan → tasks via speckit, with three decisions pinned in Clarifications before implementation:

| Decision | Outcome |
|---|---|
| Conflict rule (same cart_id exists) | **Existing server-side cart wins** + divergence notice (`restored`/`diverged` flags); the snapshot never overwrites server state |
| Key management | Shared HMAC secret across the failover network with **kid generations** (`SNAPSHOT_HMAC_KEYS="v2:...,v1:..."`); sign with current, verify current+previous, ≥24h rotation grace |
| Restore authz scope | Same tenant **and store**; any terminal within them (terminal-replacement ops covered); cross-store rejected |

**Implementation**:

- `kugel_common 0.1.54`: generic `hmac_signer` (HMAC-SHA256 over canonical JSON — always the snake_case `model_dump(mode="json")` form, so camelCase wire aliasing can't break verification; constant-time compare; kid ring parser)
- cart: `snapshot_service` (best-effort envelope build — generation failure degrades to a null field and never fails the mutation; strict verification), `signedSnapshot` attached to the 12 mutating endpoints + create (GET stays snapshot-free), `POST /api/v1/carts/restore` (verify → scope → terminal-state check → conflict → rebuild with master rehydration), audit trail collection `log_cart_restore`, error codes `4015xx`, startup key validation with degraded mode
- Failover presumes **JWT auth** (#67): terminal info is rebuilt from token claims, so a backend that has never seen the terminal can still serve the restore (the API-key path can't resolve terminal info there)
- Ops runbook: `docs/ja/cart-snapshot-key-rotation.md` (normal 24h-grace rotation / emergency rotation trade-offs)

## Tests

| Tier | Result |
|---|---|
| commons unit | 393 passed (17 new: canonicalization stability, kid generations, parse errors) |
| cart unit | 532 passed (17 new: envelope build/degrade/size-warn, every verify rejection) |
| cart integration | 46 passed (19 new: attach on all mutating endpoints, restore-into-wiped-backend → bill, master-context preservation under a master-data price change, all 4015xx rejections with audit records, replay divergence, terminal-state idempotent rejection) |
| cart e2e (live stack) | 45 passed, incl. snapshot capture → **real Redis wipe** → **JWT restore** → continue → bill, tamper rejection, existing-cart conflict |

## Measurements (T029, SC-005/SC-006)

Measured and recorded on issue #148 (c69e111): 40-item cart on the full local stack with gzip (#147) — response wire size 3.3 KB gzipped (budget 15 KB; the snapshot adds ~2.1 KB on the wire), lineItems p95 +28.2 ms vs degraded mode (budget +50 ms; p50 +2.2 ms). Both pass; no R-008 mitigation needed. Full protocol and tables in the issue comment.

## Out of scope / follow-ups

- **#152**: idempotent finalization keyed on cart_id (tranlog carries no cart_id today) — restoring a *pre-terminal* snapshot of an already-finalized transaction can still double-bill; this PR ships detection (audit trail) per spec, prevention lands there
- **#153**: pre-existing unit-test patch leak found during this work (not touched here)


kugel-masa (commented):
Self-review (/code-review) of the implementation diff. Two findings were fixed in 41fc3f1:

**🔴 Fixed — fatal Slack notification on the restore happy path**
The conflict check went through the service-level `__get_cached_cart_async`, which emits `logger.fatal` + a fatal Slack notification on *any* miss. An absent cart is the normal failover case for restore, so every successful restore would have paged ops. Now queries `cart_repo.get_cached_cart_async` directly, catches `NotFoundException` cleanly, and hydrates repositories/state inline for the existing-cart path.

**🟡 Fixed — missing terminal guard on restore**
`restore_cart_async` did not enforce the opened + signed-in terminal precondition that `create_cart_async` has. A not-signed-in terminal would have failed later with an `AttributeError` during staff injection (opaque 404/500). The same guard now runs first (404001/404002), the rejection is audited, and a regression test covers it (`test_restore_requires_opened_signed_in_terminal`, driven with the Idle/no-staff `terminal_jwt` fixture).

**🔵 Noted, not changed**
- `contracts/restore-api.yaml` omits the 422 schema-layer response (present in the endpoint definition).
- Concurrent restores of the same cart_id are last-writer-wins; benign for identical snapshots, and the audit trail records both attempts (accept + detect posture).
- `401507` is log-only by design (snapshot generation degrades, it never surfaces as an API error).

Post-fix test state: commons unit 393 / cart unit 532 / cart integration 47 / cart e2e 45 (live stack) — all green.

kugel-masa:
## Code Review Report (`/code-review`, post-implementation self-review)

### Scope

Full branch diff (`main...148-cart-snapshot-restore`): spec kit under `specs/148-cart-snapshot-restore/`, `kugel_common` 0.1.54 (`utils/hmac_signer.py`), and the cart service changes — snapshot attachment (`snapshot_service.py`, transformer, 12 mutating handlers + create), restore API (`restore_cart_async`, `POST /api/v1/carts/restore`), audit trail (`log_cart_restore`), error codes `4015xx`, settings/startup validation, tests across all three tiers, and the key-rotation runbook.

### Findings

#### 🔴 Critical — fixed in 41fc3f1

- **[Ops/Correctness]** `cart_service.py` (restore conflict check) — the existence check went through the service-level `__get_cached_cart_async`, which emits `logger.fatal` **and a fatal Slack notification on every miss**. For restore, an absent cart is the *normal* failover case, so every successful restore-into-a-fresh-backend would have paged ops.
  **Fix:** query `cart_repo.get_cached_cart_async` directly and catch `NotFoundException` cleanly; hydrate repositories/state inline on the existing-cart path; the brittle `e.original_exception` introspection is gone. Any non-not-found read error still propagates instead of risking an overwrite (FR-006).

#### 🟡 Warning — fixed in 41fc3f1

- **[Robustness/Consistency]** `restore_cart_async` lacked the opened + signed-in terminal guard that `create_cart_async` enforces. A not-signed-in terminal would fail later with an `AttributeError` during staff injection, surfacing as an opaque 404/500.
  **Fix:** the same guard now runs before any snapshot processing (`404001` terminal status / `404002` sign-in), the rejection lands in the `log_cart_restore` audit trail, and a regression test covers it (`test_restore_requires_opened_signed_in_terminal`, driven with the Idle/no-staff `terminal_jwt` fixture).

#### 🔵 Suggestions — noted, intentionally not changed

- **[Docs]** `contracts/restore-api.yaml` omits the 422 response (schema-layer rejection when required envelope fields are missing); the endpoint definition itself declares it.
- **[Concurrency]** Concurrent restores of the same `cart_id` are last-writer-wins — benign for identical snapshots; the audit trail records every attempt (consistent with the accept + detect replay posture).
- **[Style]** `401507` (snapshot generation failure) is log-only by design: generation degrades to a null `signedSnapshot` and never fails the cart operation (NFR-004).

#### ✅ Good practices confirmed

- Canonical signing form pinned to the snake_case `model_dump(mode=\"json\")` representation, so wire-level camelCase aliasing cannot break verification — validated end-to-end over a real wire round trip.
- Degraded-mode design: missing/malformed `SNAPSHOT_HMAC_KEYS` never blocks startup or cart operations; verification stays strict (all envelopes rejected).
- Audit coverage is total: restored / existing-returned / every rejection path, traceable per `cart_id` with issuing + requesting terminal recorded.
- Multi-tenancy and error-code conventions followed (tenant DB collections, cart-reserved `4015xx` block); scope checks evaluate the **signed** envelope attribution, not client-claimed values.

### Post-fix test state

| Tier | Result |
|---|---|
| commons unit | 393 passed |
| cart unit | 532 passed |
| cart integration | 47 passed (+1 terminal-guard regression) |
| cart e2e (live stack) | 45 passed — incl. Redis wipe → JWT restore → bill, tamper rejection, conflict |

### Verdict

**Approve** — findings fixed and re-verified; remaining suggestions are documentation-level and tracked in this comment.

kugel-masa:
## Second-pass review (`/code-review PR #154`) — post-merge, with an independent adversarial pass

Scope: the merge with main (gzip #149 × snapshot interaction, kugel_common 0.1.55 re-release) plus an independent adversarial review of the signing/restore core. Triage below; fixes landed in b056905.

### Fixed in b056905

- **🟡 Divergence signal polluted by staff re-injection** — the conflict path compared `staff`, which the repository overwrites from the *current* terminal info on every cache read, so a benign operator change marked restores as `diverged` and made the audit flag unreliable. `staff` is now excluded from the comparison (alongside timestamps/etag), with unit tests pinning the semantics (`test_restore_compare.py`).
- **🟡 Audit-write failure handling** — failures now always log the full audit record payload as a fallback trail; only the state-changing `restored` path still raises (previously `existing_returned` — a pure read — could 500 on an audit DB hiccup, and a failed write left no trace anywhere).
- **🔵 Signing-form drift guard** — `by_alias=False` is now pinned at the restore-request dump with a comment: the signed form is the snake_case representation, and a future `by_alias=True` would have broken every signature while looking wire-correct.

### Adversarial findings triaged as designed / already tracked

- **Replay of a pre-terminal snapshot after finalization (raised as critical)** — this is the documented phase-1 posture (accept + detect; spec Clarifications 2026-06-12) and is tracked in **#152** (cart_id-keyed idempotent finalization). One correction to the finding: the snapshot carries `transaction_no=-1` (numbering happens at bill), so a replayed bill takes a **new** number — the risk is a duplicate sale, not a key collision. Detection (audit trail) ships in this PR; prevention is #152.
- **"Restore bypasses the state machine" (raised as critical)** — rejected: restore re-materializes a state the server itself computed and signed (FR-010); it is constrained to `RESTORABLE_STATUSES` (Idle/EnteringItem/Paying), and restoring into Paying and then billing is precisely the failover-continuation use case. `check_event_sequence` guards *events*; restore is not an event.
- **No `issued_at` freshness window** — deliberate spec decision for phase 1 (no TTL/nonce); worth revisiting together with #152.
- **Extra envelope fields silently dropped** — confirmed harmless for the signature (the signed payload is exactly the declared fields both at issue and verify); kept permissive intentionally for rolling-upgrade tolerance, with `schema_version` governing compatibility.
- Key handling confirmed clean: no key material in logs or error messages; constant-time comparison; degraded mode rejects all envelopes.

### New observation (follow-up candidate, not in this PR)

- **🟡 Request-log amplification** — `log_requests` middleware stores the **full response body** in both the log file and the `request_log` collection. With snapshots attached, every mutating response body roughly doubles (cart + full snapshot incl. masters), so request-log storage grows accordingly across file + Mongo. Mitigation belongs in the shared middleware (truncate bodies or strip `signedSnapshot` from logged bodies) and should be sized together with T029 measurements. Can be filed as a separate issue.

### Post-fix test state

commons unit 397 / cart unit **536** / cart integration 47 — green. (One environment note: a stale `db_cart_test_tenant` DB in the local Mongo volume collided case-insensitively with `db_cart_TEST_TENANT` in the rollover tests; dropped — not a code issue.)

### Verdict

**Approve** — protocol-level replay prevention remains explicitly scoped to #152 per spec; everything actionable in this PR is fixed and verified.

PR #157, opened 2026-06-14
feat: client-carried cart phase 2 — stateless cart backend (#156)

### LINKED ISSUE #156: cart: carry the cart snapshot in every mutating request — stateless cart backend (client-carried cart phase 2)
Phase 2 of the #146 client-carried cart design. Phase 1 (#148, merged in #154) made every cart-mutating response carry a signed snapshot and added a restore API. #146 gated phase 2 on the phase 1 measurements; those are in (issue #148, T029): 3.3 KB gzipped for a 40-item cart response (budget 15 KB) and lineItems p95 +28.2 ms (budget +50 ms), so phase 2 is a go.

## Goal

The POS carries the cart in **every request**, not just across failovers: each cart-mutating request presents the latest signed snapshot, the backend verifies and re-materializes the cart from it, applies the operation, and returns a new snapshot. The server-side cart cache (Redis / MongoDB `cache_cart` fallback) is demoted from authority — the client-carried snapshot becomes the source of truth, and the timeout→404 / split-brain / empty-fallback problem class disappears entirely (see #146 for the analysis).

## Scope (to be refined by spec)

- Cart-mutating APIs accept the signed snapshot envelope in the request (existing-cart operations); verification reuses the phase 1 signer/kid-ring as-is
- Server-side cache demoted from authority: kept as an optional optimization or removed (open question — decide in spec)
- Request-body decompression middleware with a decompressed-size guard (the deferred work noted in `kugel_common` `http_compression.py`), so clients can send compressed envelopes
- Replay posture stays accept + detect (#146): stale-snapshot resubmission becomes normal operation; divergence/staleness must remain auditable
- Compatibility story for phase-1-style clients that do not send snapshots (open question — decide in spec)

## Dependencies

- **#152 (idempotent finalization keyed on cart_id)** becomes a hard prerequisite for double-count prevention once the server no longer holds authoritative cart state
- #148 / #154 (phase 1, merged), #67 (JWT terminal-info reconstruction), #147 (gzip)

Spec kit to follow in `specs/` on a branch named after this issue.

kugel-masa:
Closes #156. Phase 2 of the #146 client-carried cart design: the POS carries the signed cart snapshot in **every mutating request**, the backend reconstructs the cart from it and applies the operation statelessly, and the server-side cache is demoted from authority. This eliminates the cart-cache failure class (Redis timeout → 404, split-brain, empty fallback) and makes finalize deterministic + idempotent so duplicate finalizes converge to one transaction.

Built spec-kit first (`specs/156-stateless-cart/`: spec → plan → research → data-model → contracts → tasks), with the load-bearing design decisions pinned in Clarifications before implementation.

## Key decisions (Clarifications)

| Decision | Outcome |
|---|---|
| Request transport | Endpoints take top-level JSON arrays, so a sibling field is impossible. Phase 2 clients send a **wrapped body `{signedSnapshot, payload}`**; a peel middleware lifts the snapshot to `request.state` and forwards `payload` unchanged (endpoint signatures untouched, DI branches, backward compatible). |
| Migration | **Dual-mode** via a service-wide env (`DUAL` accepts the no-snapshot/cache path; `REQUIRED` rejects it). Cache removal deferred to post-migration. |
| Divergence | Under 1-cart-1-client, forks are anomalies → the **presented snapshot is authoritative**, divergence is audited not resolved. |
| Numbering | `transaction_no` becomes a carried `(business_counter, seq)` composite; `receipt_no` stays a **continuous** counter; the transaction **time is client-stamped at bill**. All carried so a retried finalize on any backend yields the same number/time/receipt. |
| Offline open | `business_counter` and `receipt_no` are **terminal-owned**; terminal service is the durable home and **reconciles via max()** at open so an offline-used value is never reused (gaps allowed). |
| Double-count guard | No finalize gate. `cart_id` rides the tranlog; downstream **report/journal/stock dedupe on cart_id (first-wins skip)** so a duplicate finalize converges to one record. This is #152's mechanism. |

## What changed

- **receipt**: the sales receipt now prints the full transaction identity (store code / register / 営業回 / 取引通番). It previously showed only register and receipt number, which no longer names one sale — `transaction_no` restarts every open — so a return presented at another store had nothing to look the original up by. The return receipt's original block gained 営業回 for the same reason.
- **cart**: wrapped-request peel middleware; DI branch + `prepare_stateless_from_snapshot` (verify → scope/state → reconstruct, cache-independent, FR-004); carried `(business_counter, seq)` + client-stamped time at bill via a `FinalizeContext`; `seq`/`transaction_datetime` on the cart; `api_path` in the audit trail; `CART_REQUEST_SNAPSHOT_MODE`. The carried snapshot must name the same cart as the URL (401512) — otherwise the reconstructed cart would silently replace the addressed one. Transaction lookup (get / void / return) takes the open epoch, since `transaction_no` is now the per-open seq: an optional `business_counter` query parameter, and an ambiguous match without it is refused (409, 401513) rather than guessed at. `transaction_status` gained the same epoch in its identity and unique index — a daily open resets seq to 1, so day 2's first sale was reading and overwriting day 1's void/refund status.
- **commons (0.1.63)**: `RequestDecompressionMiddleware` — gzip / deflate / br request bodies (FR-009), with the decompressed-size ceiling enforced during expansion so a forged small body cannot exhaust memory (413 + 401509). Registered outermost in cart, since the snapshot peel JSON-parses the body and a still-compressed one would silently read as a legacy snapshot-less request. An encoding it cannot expand (`zstd`, or a chained `gzip, br`) is refused with 415 rather than passed through, for the same reason. `cart_id` on `BaseTransaction`; `receipt_no` on `TerminalInfoDocument` + token claim + reconstruction.
- **report / journal / stock**: dedupe finalize on `cart_id` (first-wins skip); tran/stock_update unique indexes reworked (partial `cart_id`, `business_counter` in the numbering tuple). Stock's `$inc`+rollback race backstop preserved.
- **terminal**: `open_terminal_async` max-reconciles client-carried `business_counter`/`receipt_no`; `TerminalOpenRequest` carries them (bounded `ge=0` + ceiling, plus a jump guard against the stored value — the reconcile is monotonic and therefore irreversible); `receipt_no` initialized on first open.
- All 7 services aligned on kugel_common 0.1.63 (rebased onto main after #160).

## Breaking changes

Phase 1 (#148/#154) endpoints retired now that the snapshot rides every request:

| Removed | Replacement |
|---|---|
| `POST /api/v1/carts/{cart_id}/restore` | Carry the snapshot on the operation itself — a wrapped request reconstructs the cart, so there is nothing to restore first. |
| `GET /api/v1/carts/{cart_id}` | The client already holds the cart in its snapshot; the server no longer serves it as an authority. |

DUAL mode keeps snapshot-less **mutating** requests working during the rollout, but it does not bring these two endpoints back. A phase 1 client that calls either must move to the wrapped-request flow.

Behaviour changes on the transaction endpoints:

- **Void is now confined to the terminal's current business date AND open session** (400, code 401514). It was already limited to this terminal, but nothing checked the date or epoch — a past sale could be voided outright, prevented only when its `transaction_no` happened to be ambiguous. A void reverses a sale while the drawer and the day's totals are still open; past sales are handled by return, which books its own transaction instead of editing a settled day.

- **Returns accept an original from any store or terminal of the tenant.** The customer brings the receipt to whichever store they choose, and the receipt carries `store_code + terminal_no + business_counter + transaction_no`. The path `store_code`/`terminal_no` name where the ORIGINAL was rung up and no longer have to match the authenticated terminal; the tenant boundary is still enforced. The return itself is booked against the terminal performing it — previously it inherited the original's store/terminal while stamping this terminal's counters.
- **`transaction_no` alone may be ambiguous.** It is the per-open seq now, so clients should send `business_counter`. Omitting it still works while the number resolves to one transaction (server-numbered, pre-phase-2 records), and is rejected with 409 once it does not.

## Tests

| Tier | Result |
|---|---|
| cart | unit 583, integration 46 (per-request roundtrip after cache wipe, tamper rejection + audit api_path, carried numbering, cross-cart snapshot rejection, epoch lookup + ambiguity refusal, cross-store return) |
| report / journal / stock | integration 47 / 14 / 35 (cart_id dedupe; no regression) |
| commons | unit +15 (expansion incl. chunked bodies, zip-bomb refusal at and past the ceiling, unsupported/chained/identity encodings, brotli-absent degradation, header rewrite) |
| terminal | unit 345, integration 15 (open reconciles client-carried counters into the token; implausible/negative carried values rejected) |
| **e2e (live full stack)** | 8 per-service suites (cart 56, incl. a void/return reach suite: void refused for a past session / past business date / another terminal and accepted for the current one, return accepted across stores from an old session with the persisted origin and refund record verified, tenant still refused, ambiguous number refused; incl. a new stateless/compressed suite: wrapped request over the wire, gzipped, a body large enough to arrive split across ASGI messages, and the 413/415 guards) + cross-service 31 (pubsub idempotency, data consistency) + a **phase-2 stateless-chain e2e**: wipe the cart from live Redis, then drive add/subtotal/pay/bill entirely through wrapped requests; carried finalize context drives numbering. All pass, no regression. |

## Index migration (runs on tenant creation, not at startup)

`create_collection_with_indexes_async` drops the stale key pattern, ensures the new ones, and then **verifies the end state and raises `DatabaseException` if it was not reached** — a silently skipped step fails open (a stale unique index blocking finalize inserts, or a missing `cart_id` dedupe index) and looks healthy. It handles an existing collection correctly, covering the tranlog / stock_update indexes and cart's `transaction_status`, whose data backfill runs first.

**Its only trigger is the tenant creation API — it does not run at application startup.** A new tenant is migrated on creation; an existing one is not. Nothing is affected today (there is no production data yet, and the test suites drop and recreate their tenant through that API, which is why they pass), but **introducing this into a deployment that already has tenants needs a deliberate way to run the migration over them**. An earlier revision of this description claimed it ran at startup; that was wrong.

`transaction_status` additionally gets a **data backfill** before its new unique key is enforced: existing records take `business_counter` from the transaction they refer to. Matching them with "epoch or null" instead is not viable — legacy `transaction_no` came from a 1-based per-terminal counter and `seq` is 1-based per open session, so the ranges overlap completely and a lookup for this session's `seq=1` would find the terminal's very first sale. A record whose transaction cannot be resolved unambiguously is left alone and logged. A silently-skipped drop or ensure would otherwise fail open — a stale unique index blocking finalize inserts, or the new `cart_id` dedupe index missing — and look healthy. Startup fails loudly instead, so plan the rollout accordingly.

## Deployment note

`SNAPSHOT_HMAC_KEYS` was previously set nowhere — not in compose, not in `.env.sample` — so the running stack had the snapshot feature **degraded since phase 1**: no snapshot issued, every carried one rejected, clients silently falling back to the server-side cache. Development compose supplies a working default so the e2e suite can exercise the stateless path, and `.env.sample` documents the setting (with `REQUEST_DECOMPRESS_MAX_BYTES` and `CART_REQUEST_SNAPSHOT_MODE`) but **ships no key** — only the generator command, since README tells operators to `cp .env.sample .env`. The compose default is public by definition, so `init_snapshot_signer` recognises the key material committed here and logs it at **ERROR** at startup: unlike a missing key it loads and verifies, so every other signal looks healthy while the signature is worthless. **`docker-compose.prod.yaml` deliberately carries no default: production must supply its own key**, or phase 2 clients degrade silently. The first e2e test now asserts a snapshot is actually issued, which is the check that would have caught this.

Signing keys are read once at startup, so rotating them implies a restart. The runbook now rotates in **two phases** — distribute the new key for acceptance only, confirm every backend has it, then reorder the ring to switch signing — because edge and cloud backends cannot be updated simultaneously, and a single-step rollout breaks one direction while in progress: an updated backend signs with a key the un-updated ones cannot verify, so a transaction succeeds or fails depending on where it fails over to.

## Documentation

The published specs described a system this branch no longer implements, so they were brought back in line.

- **API specifications regenerated** (`docs/{ja,en}/*/api-specification.md`, `docs/openapi/*.json`). The cart spec still listed `GET /carts/{cart_id}` and the restore endpoint — both retired by FR-010 — and knew nothing of the `business_counter` query parameter. Regenerating also picked up master-data and terminal drift accumulated since the last run.
- **New: `docs/{ja,en}/cart/client-carried-cart.md`.** None of phase 2 was written down anywhere: the wrapped request shape, compressed bodies and their ceiling, DUAL vs REQUIRED, `transaction_no` becoming the per-open seq (identity is the `(business_counter, transaction_no)` pair), carried counters and their bounds, why void reaches only the current session while return reaches any store and any past session, the 4015xx error codes with their HTTP statuses, `cart_id`-keyed idempotency, the index migration, its trigger, and why it deliberately aborts on failure, and the removed endpoints.
- **Key rotation runbook** (`docs/ja/cart-snapshot-key-rotation.md`) was written against the restore API and described monitoring "restore attempts"; verification now happens on the per-request path. Added the two operational facts it lacked: an unset key degrades **silently** into the cache path, and the new ERROR that fires when a key published in this repository is in use — the one case where signatures verify and protect nothing.
- **Model specification** presented the Dapr State Store as the cart's primary storage, which is precisely what phase 2 stopped being true.

Two fixes were needed along the way, both pre-existing:

- `scripts/export-openapi.sh` exported exactly one service and stopped. Under `set -e`, `((SUCCESS_COUNT++))` evaluates to the old value, so the first increment returns 0 — a non-zero status that ends the script. All seven export now.
- Four broken links in both READMEs (the files are `design_patterns.md` / `http_communication.md`; the links used hyphens). `docs/` now has none.

Out of scope: the external Zenn articles under a separate repository still describe the cart cache as authoritative. That framing only becomes wrong once the cache is removed, which is a later phase.

## Out of scope / follow-ups

- **#152**: this PR implements its mechanism (cart_id dedupe downstream); the issue tracks the broader idempotent-finalization story.
- receipt_no configurable start/rollover, cancel/void carried numbering, audit collection rename (cosmetic).
- Clients that omit `business_counter` on **return** keep working only while the number still resolves to one transaction; from the second open session onward they must send it (409 otherwise). **Void** needs no change: it reaches only the terminal's current session, so an omitted epoch defaults to it.
- Cache/circuit-breaker removal lands after migration (dual-mode keeps them during the rollout).


kugel-masa:
## Performance verification: legacy vs stateless

Full write-up committed at [`specs/156-stateless-cart/performance.md`](../blob/156-stateless-cart/specs/156-stateless-cart/performance.md). Measured on the live docker stack (DUAL mode, single client, warm), continue-ops `subtotal`/`payment`/`bill` across basket sizes N∈{1,5,20,50} with **unique item codes** for realistic compression.

**Latency (mean ms)** — on par between patterns; stateless slightly ahead on large baskets (cache bypass + carried numbering skip the Redis/counter round-trips):

| N | op | legacy | stateless | Δ |
|--:|----|------:|---------:|--:|
| 5 | bill | 30.6 | 31.0 | +0.4 |
| 50 | subtotal | 27.4 | 21.0 | −6.4 |
| 50 | bill | 56.5 | 51.0 | −5.5 |

Latency grows with basket size for both; sub-10ms deltas are within M=10–12 noise.

**Transport (carried snapshot, on-wire):** request-direction only — response is identical since DUAL mode returns a snapshot on both paths.

| N | raw | gzip | brotli |
|--:|----:|-----:|-------:|
| 1 | 5.4 KB | 1.31 KB | 1.21 KB |
| 50 | 50.4 KB | 3.64 KB | 2.93 KB |

≈ **1.3 KB + ~40 B/line item compressed**; a 50-line basket ≈ 3 KB (brotli), matching the issue #148 ~3.3 KB figure.

**Takeaway:** stateless achieves the cache-loss-resilient / any-backend-continuation goal with no meaningful latency penalty; the only cost is request-direction transport (a few KB compressed). Caveats (single-client, warm, loopback, no concurrency) are in the doc.

kugel-masa:
### Correction/addendum: addItem is the dominant factor

The first write-up omitted `addItem` — the per-scan op whose **response carries the full cart + signed snapshot**. Added to [`performance.md`](../blob/156-stateless-cart/specs/156-stateless-cart/performance.md):

**addItem latency (mean ms)** — stateless is consistently faster and the gap widens with basket size (it skips the Redis read+write of the growing cart):

| cartItems | legacy | stateless | Δ |
|--:|------:|---------:|--:|
| 1 | 19.3 | 16.0 | −3.2 |
| 20 | 24.7 | 19.2 | −5.5 |
| 50 | 35.2 | 28.5 | −6.7 |

**addItem response size** (identical for both patterns under DUAL — the snapshot is returned regardless):

| cartItems | raw | gzip | brotli | snapshot % of body |
|--:|----:|-----:|-------:|--:|
| 1 | 6.6 KB | 1.60 KB | 1.49 KB | 80% |
| 50 | 66.5 KB | 4.97 KB | 3.94 KB | 76% |

**The snapshot is ~76–80% of every addItem response**, returned on every scan, and the addItem response is ~4–5× a phase-1 cart-only response. This — not the request-direction snapshot — is the dominant transport cost of phase 2, and it largely duplicates the cart view already in the response (dedup / delta-snapshot / separate-fetch are the optimization surface). Latency is never worse for stateless; the cost is purely response-direction bytes.

kugel-masa:
### Cold-cache (post-restart) measurement

`stop.sh`/`start.sh` brings up a **fresh empty Redis** (no persistence volume), so a restart clears the cart + master-data caches (MongoDB data survives). Measured the post-restart transient with **no warmup** (N=10, 12 iters, separate item-code ranges per pattern). Added as §4.5 of [`performance.md`](../blob/156-stateless-cart/specs/156-stateless-cart/performance.md).

**cold (iter 1–2) vs warm (last 4), mean ms:**

| metric | legacy cold | legacy warm | stateless cold | stateless warm |
|--------|----:|----:|----:|----:|
| firstScan (1st addItem) | 30.9 | 29.4 | 39.5 | 25.1 |
| bill | 64.3 | 56.5 | 33.7 | 32.3 |
| total tx | 450.4 | 388.1 | 401.1 | 352.8 |

Coldest single iter1: legacy bill **77.1ms** vs stateless bill **35.4ms**.

**Findings:**
- **The cold penalty is larger for legacy.** Its `bill` does server-side counter + settings numbering that is expensive on cold caches/pools (iter1 77ms); stateless carried numbering sidesteps it (iter1 35ms) — up to ~42ms gap cold.
- **Stateless's only cold cost is the first `addItem`** (snapshot machinery + first-request serialization warming up, ~55ms iter1); it converges within ~2 iterations and is then faster than legacy.
- **Full transaction is faster with stateless both cold and warm.**

This corroborates the earlier note that the one-off "−22ms bill" reading was a cold artifact — and shows the cold legacy-bill penalty recurs on every cold start, which stateless avoids.

kugel-masa:
## Follow-up: void/return carried numbering + deterministic DailyInfo fingerprint

Commit `4174966` addresses two gaps found while reviewing how the new `(business_counter, seq)` numbering interacts with downstream verification. Both are public-scoped (no offline-settlement work — that stays in the edge/ST-POS layer).

### Problem 1 — void/return used a different numbering space than sales

After this branch, **sales** draw `transaction_no` from the terminal-carried per-open `seq` (1, 2, 3…), but **void/return** still used the server-side continuous counter:

```
one open session:
  sale  -> seq = 1
  sale  -> seq = 2
  void  -> server counter = 5001   <- different space
```

`(business_counter, transaction_no)` is a **unique index**, so on a fresh terminal a `void` numbered `1` could collide with `sale seq=1` and fail to insert. The sequence is also no longer a clean per-open run.

**Fix:** void/return now draw their `seq` / `receipt_no` / time from the **same per-open carried sequence as sales**, so all transaction types share one numbering space within a session.

### Problem 2 — the all-arrived verification fingerprint was time-ordered

The terminal close-log (`cart_transaction_last_no`) and the report `DailyInfo` verification both picked the "last transaction" by `generate_date_time` desc. Since #156 makes that timestamp **client-stamped**, two finalizes in one session can tie — and the two independent queries (terminal vs report) could then pick **different** records, causing a spurious `TransactionMissingException` even though every log actually arrived.

**Fix:** both sides now order by `(business_counter, transaction_no)` — the canonical unique key. No ties, so both sides deterministically pick the same record. (Depends on Problem 1's unified numbering.)

### How void/return carry the numbers — signed finalize-context envelope (Option B)

void/return bodies are a bare `list[PaymentRequest]`, so the numbers can't go in the payload. Instead of query params (forgeable, no idempotency), they ride a **signed finalize-context envelope** in the existing `signedSnapshot` wrapper:

```json
{ "signedSnapshot": <finalize-context envelope>, "payload": [ <payments> ] }
```

- The peel middleware already runs on void/return, so the **HTTP contract is unchanged** (path / body shape / response).
- The envelope carries `{ cart_id, seq, receipt_no, transaction_datetime }`, HMAC-signed (numbers can't be forged — NFR-003). It is a *finalize context*, not a cart snapshot (no cart reconstruction; scope is checked against the terminal).
- **DUAL mode:** no envelope → legacy server numbering + fresh `cart_id` (existing clients unaffected).

### Bonus: void/return are now retry-idempotent

Previously void/return minted a fresh random `cart_id` per call, so a lost-ACK retry created a **second void → double inventory reversal**. The envelope carries a **stable `cart_id`**, so a retry converges via the downstream `cart_id` first-wins dedupe → exactly one void.

```
before: void retry -> cart_id=BBB -> 2nd void created    (double reversal)
after:  void retry -> cart_id=AAA -> already processed, skipped (idempotent)
```

### Interface impact

None for existing clients. New HTTP request/response fields: zero. New Python params are optional with defaults. A 100%-legacy deployment behaves identically; carried numbering only activates when a client opts in via the signed envelope.

### Tests

- cart unit **568**, terminal **345**, report **223** (+11 new; fixed 2 stale report/journal dedup-filter tests that hadn't picked up `business_counter`)
- integration **7/7** services (+2 new: carried void uses signed context; tampered seq rejected)
- e2e **8/8** targets, incl. `test_void_return_journey` on the live Dapr stack

### Spec

`specs/156-stateless-cart/` updated to match: Clarifications 2026-06-15, FR-012 / FR-013, research R-011 / R-012, contract `FinalizeContextEnvelope`, data-model §7/§8.

kugel-masa:
## Performance — snapshot-absent vs snapshot-carried (Locust A/B)

Added a Locust A/B harness (`performance_tests/locustfile_snapshot_compare.py`, run via `run_perf_test.sh … --compare`) that drives **both cart paths in one run** over the same stack / terminals / load, so they're directly comparable:

- **legacy** — bare request bodies → cache-authoritative path
- **stateless** — every mutating request carries the prior response's `signedSnapshot` in a wrapped `{signedSnapshot, payload}` body → reconstruct-from-snapshot, no cache read

Request names are tagged `[legacy]` / `[stateless]`; a delta report (`generate_snapshot_compare_report.py`) pairs them per op.

### LAN / loopback — 40 users, 20 items/cart, 5 min (3,947 reqs, **0 failures**)

| Operation | legacy mean | stateless mean | Δ | p95 L→S |
|---|---|---|---|---|
| Create Cart | 428 ms | 406 ms | −5.2% | 1100→1200 |
| Add Item (hot path) | 147 ms | **132 ms** | **−10.7%** | 500→450 |
| Cancel Cart | 2653 ms | 2660 ms | +0.2% | 6200→5800 |

Stateless is **as-fast-or-faster**: it rebuilds the cart from the carried snapshot and skips the Redis read+write, and the gain **grows with cart size** (more cache I/O avoided). Consistent with `specs/156-stateless-cart/performance.md`.

### Remote backend / constrained upload — `tc netem`, upload-only throttle (directional)

The stateless path uploads the (growing) snapshot on **every** request, so on a slow/asymmetric uplink that cost can overtake the *server-side, local* Redis round-trip it avoids → stateless turns slower:

- stateless slower at **256 kbit** upload (mean +27%); at **1 mbit** the stateless **p95 blew up to 9100 ms vs 3900** — the penalty hits the tail first.
- ≥ ~512 kbit: stateless faster or even.
- **RTT is not the factor** (round-trip count is unchanged) — **low upload bandwidth** is.
- ⚠️ Measured with **uncompressed** ~14 KB snapshots; FR-009 compression (~3–5 KB) pushes the crossover **~3–4× lower** in bandwidth.
- This single-user micro-benchmark is **too noisy to pin the exact crossover** (non-monotonic points); treat as directional.

**Takeaway:** stateless trades server-side cart state for per-request upload bandwidth. Net win when the client↔backend link is fast (edge proximity / LAN); only on genuinely low-upload remote links does it lose, and compression mitigates.

### Reproduce
```bash
cd services/cart/performance_tests
./scripts/run_perf_test.sh setup 50
./scripts/run_perf_test.sh custom 40 5m --compare   # [legacy] vs [stateless], one run
```

kugel-masa (commented):
## Review summary — ready to merge

Reviewed across several passes as the branch evolved. Everything raised has been addressed; recording the outcome here since GitHub does not allow approving one's own PR.

### Defects found and fixed during review

| Severity | Finding | Fix |
|---|---|---|
| 🔴 | `.env.sample` shipped a real signing key, and README instructs `cp .env.sample .env` — the documented setup path would have deployed a key published in this repository. Anyone reading it could sign a snapshot with arbitrary prices. | Sample ships no key. `init_snapshot_signer` now recognises repo-committed key material and logs it at **ERROR**, since unlike a missing key it loads and verifies while protecting nothing. |
| 🔴 | `transaction_status` matched "epoch or null" to reach pre-migration rows. Legacy `transaction_no` and `seq` are both 1-based, so the ranges overlap completely — verified against MongoDB that a lookup for this session's `seq=1` returns the terminal's first-ever sale. A void would have been refused as "already voided" constantly. | Exact match plus a startup backfill that takes each row's epoch from the transaction it refers to. |
| 🟡 | The carried snapshot could name a different cart than the URL; the reconstructed cart silently replaced the addressed one, invisible even in the audit trail. | Rejected with a dedicated code (401512). |
| 🟡 | `business_counter` / `receipt_no` arrived unbounded and are reconciled with a monotonic `max()`, so one malformed value would permanently burn the terminal's number space. | Bounded in the schema plus a jump guard against the stored value. |
| 🟡 | Void was limited to this terminal but **not to today** — nothing checked business date or epoch anywhere. A past sale could be voided; the only thing sometimes preventing it was whether its number happened to be ambiguous. | Confined to the current business date and open session (401514), pointing at return. |
| 🟡 | Returns are booked against the performing terminal, but `return_async` left the original's store/terminal in place while stamping this terminal's counters — a numbering tuple belonging to neither. | Attribution corrected; `origin.business_counter` recorded so the original stays identifiable. |
| 🟡 | The sales receipt printed only register and receipt number. Since `transaction_no` restarts every open, it named nothing — a return presented at another store had no way to look the original up. | Prints store code / register / 営業回数 / 取引通番. |
| 🟡 | Reporting 404 for a sale older than this session told the cashier the receipt number was wrong, so they would retype instead of switching to a return. | Distinguishes "no such number anywhere" (404) from "out of a void's reach" (401514). |

### Also closed

- **FR-009 was unimplemented.** `REQUEST_DECOMPRESS_MAX_BYTES` existed but nothing read it, and a client that did compress hit the snapshot peel, failed to JSON-parse, and silently took the cache-authoritative path. Now expanded outside the peel, with the ceiling enforced during expansion; unexpandable encodings refused rather than passed through.
- **`SNAPSHOT_HMAC_KEYS` was configured nowhere**, so the stack had run degraded since phase 1 and the stateless path had never executed end to end.
- **ruff was unusable as documented** — only black was configured, so `ruff format` per CLAUDE.md would have rewritten the repository.

### Verification

Unit 8/8 services · integration 7/7 · e2e 8/8 · cross-service e2e 31/31, all against the live stack. Behaviour that could not be reproduced in-process — compressed and chunked request bodies, cross-store returns, void scope boundaries — was checked over the wire rather than through the ASGI harness.

### Before deploying

**Set `SNAPSHOT_HMAC_KEYS` in production.** Unset, phase 2 degrades silently to the server-side cache; inherited from the development compose, signatures become worthless. Both are now visible in the startup log, but neither stops the service.

Clients omitting `business_counter` on **return** must start sending it from the second open session onward. **Void** needs no client change.

### Not done

342 files carry the Apache header collapsed onto one line (38 touched here). It predates this PR and wants its own mechanical commit rather than an inconsistent partial fix buried in this diff.

Issue #158, opened 2026-08-12
infra: unify inter-service calls on Dapr service invocation (retire the direct-HTTP compose override)


kugel-masa:
## Goal

Make Dapr service invocation the single transport for **HTTP** inter-service calls, and retire the direct-HTTP configuration that currently shadows it. This is the upstream counterpart of the fork-side discussion in `retail-ai-inc/stpos-backend-kugelpos#951`, where the symptom surfaced as a missing `account` sidecar in the cloud profile.

## The split is configuration, not code

Every inter-service HTTP call already goes through `HttpClientHelper` (`get_service_client` / `get_pooled_client`, resolved via `_get_service_url` in `services/commons/src/kugel_common/utils/http_client_helper.py:351`). Whether a call travels direct or via a sidecar is decided purely by the value of `BASE_URL_*`. The call sites do not change, and `architecture-principles.md` ("inter-service HTTP must go through `HttpClientHelper`") stays as-is.

Today three configuration sources disagree:

| Source | `BASE_URL_MASTER_DATA` | Transport |
|---|---|---|
| `services/*/.env.sample` | `${BASE_URL_DAPR}/invoke/master-data/method/api/v1` | invoke |
| `services/docker-compose.yaml` | `http://localhost:3500/v1.0/invoke/master-data/method/api/v1` | invoke |
| `services/docker-compose.override.yaml` | `http://master-data:8000/api/v1` | **direct** |
| `services/docker-compose.prod.yaml` | `http://master-data:8000/api/v1` | **direct** |

`docker-compose.override.yaml` is applied by default (both by filename and explicitly via `scripts/start.sh:18`), so the base file's invoke configuration is silently reverted in every local run. The direct-HTTP block has been there since the initial commit, labelled "Daprを使わない場合の直接サービス間通信設定" — most likely for debug-attach runs. `debugpy` is still a dependency in all seven `Pipfile`s and `.vscode/launch.json` is a debugpy attach config on port 5678, but no compose file or script launches a service under `debugpy` anymore. The mode is vestigial.

## gRPC is a validated path, not an exception to clean up

`docker-compose.prod.yaml` sets `USE_GRPC: ${USE_GRPC:-true}` for both cart and master-data, with `MASTER_DATA_GRPC_URL: master-data:50051` (`:188`, `:243-245`). The hottest inter-service call — cart's item-master lookup — therefore does **not** travel over HTTP at all in the measured configuration. This shapes the migration:

[...]

PR #160, opened 2026-08-13
infra: align inter-service transport on direct HTTP and close the unset BASE_URL_* defect class

### LINKED ISSUE #159: infra: align inter-service transport on direct HTTP and close the unset BASE_URL_* defect class
Follow-up to #158, which was closed after the transport policy was reversed.

## Policy

**Local (in-store / edge, docker-compose) deployments use direct HTTP and gRPC for inter-service calls.** Dapr is not on the path of a sale: a daprd outage must not be able to stop front-of-store operations. Dapr remains in use for pub/sub, state stores, and bindings, which are asynchronous and guarded by the circuit breaker in `DaprClientHelper`.

See #158 for the full rationale and for lower-priority findings not covered here.

## Status

All items are complete. Landed on `bugfix/base-url-omissions`:

| Commit | Scope |
|---|---|
| `f4a97b2` | 1 — the six omissions |
| `8694d34` | 3 — `_get_service_url` hardening, demo block removal |
| `739b328` | 3 — startup verification, `.env.sample` conversion, READMEs |
| `abe3605` | 2 — override folded into the base file and deleted |
| `1b22dc1` | integration conftests resolve respx bases from `settings` |
| `6f4c8a7` | 3 — contract test across code, declaration and configuration |

The contract test found an over-declaration on its first run: `stock` listed
`BASE_URL_MASTER_DATA` but never resolves it. That declaration had been derived
from what the compose files happened to set rather than from the call sites —
the same backwards reasoning that produced the wrong requirement table in this
issue originally. Corrected to `BASE_URL_TERMINAL` and `BASE_URL_CART`.

Verified end to end: 1911 unit, 172 integration, 8 e2e suites (including the
cross-service scenarios), all green. The stack was rebuilt and restarted from the
merged base file alone; all seven services report healthy and the six with
declared requirements log `service URL settings verified` from inside their
containers. `grep -r invoke services/docker-compose*.yaml services/*/.env.sample`
returns nothing, and no reference to `docker-compose.override.yaml` remains
anywhere in the repository.

## Problem

Four independent copies of the same inter-service wiring exist, and they disagree:

| Source | Transport | Matches policy? |
|---|---|---|
| `services/docker-compose.override.yaml` | direct | yes |
| `services/docker-compose.prod.yaml` | direct | yes |
| `services/docker-compose.yaml` | invoke | **no** |
| `services/*/.env.sample` | invoke | **no** |

Duplicating the wiring four ways has already produced the same class of defect **six times** — a `BASE_URL_*` a service actually uses is simply absent, so it silently falls back to a `settings_web.py` default that resolves to nothing inside the container.

### Known omissions

Audited against actual call sites. `BASE_URL_TERMINAL` is required by every API-key-authenticating service because `get_terminal_info_from_terminal_service` (`services/commons/src/kugel_common/security.py:302`) calls `get_pooled_client("terminal")` from commons.

Note on auditing: call sites use both the positional form `get_service_client("cart")` and the keyword form `get_service_client(service_name="cart")`. A grep that only matches the positional form under-reports — that is how items 5 and 6 below were initially missed. Any check written for this must cover both.

1. **`report` is missing `BASE_URL_MASTER_DATA` in three places** — `docker-compose.yaml`, `docker-compose.prod.yaml`, and `services/report/.env.sample`. `report` reads it in `item_report_maker.py:50,54`, `category_report_maker.py:49`, and `payment_report_maker.py:200`; it falls back to the `settings_web.py:39` default `http://localhost:8002/api/v1`, which resolves to nothing inside the report container. In the base compose file the defect is currently masked by `docker-compose.override.yaml`. In the prod file nothing masks it, though report has never been exercised on that profile.

2. **`terminal` is missing `BASE_URL_STOCK` in `services/terminal/.env.sample`** — used at `services/terminal/app/api/v1/tenant.py:87`. Present in both compose files, absent from the sample.

3. **`TOKEN_URL` is unset for all six non-account services in `docker-compose.prod.yaml`** — falls back to `settings_auth.py:38` (`http://localhost:8000/api/v1/accounts/token`), which under `network_mode: service:<svc>` points at the service's own app port. Runtime impact is limited to the OpenAPI Authorize URL (`security.py:37`), but the base compose file sets it for all six, so this is a plain omission.

4. **`services/stock/.env.sample` does not exist**, while stock requires `BASE_URL_TERMINAL`, `BASE_URL_MASTER_DATA`, and `BASE_URL_CART`.

5. **`report` is missing `BASE_URL_CART` in `services/report/.env.sample`** — used at `services/report/app/api/v1/tran.py:293` (`get_service_client(service_name="cart")`). Present in both compose files.

6. **`journal` is missing `BASE_URL_CART` in `services/journal/.env.sample`** — used at `services/journal/app/api/v1/tran.py:288`. Present in both compose files.

### Required `BASE_URL_*` per service

| Service | Required by code |
|---|---|
| account | — |
| terminal | MASTER_DATA, CART, REPORT, JOURNAL, STOCK, TERMINAL |
| master-data | TERMINAL |
| cart | TERMINAL, MASTER_DATA, DAPR |
| report | TERMINAL, MASTER_DATA, JOURNAL, CART |
| journal | TERMINAL, CART |
| stock | TERMINAL, MASTER_DATA, CART |

`BASE_URL_DAPR` stays as-is everywhere — it addresses the sidecar for state and pub/sub (`services/cart/app/models/repositories/cart_repository.py:49`), not inter-service calls.

### 1. Fix the omissions

- [x] Add `BASE_URL_MASTER_DATA` to `report` in `docker-compose.yaml`, `docker-compose.prod.yaml`, and `services/report/.env.sample` — done in f4a97b2
- [x] Add `BASE_URL_STOCK` to `services/terminal/.env.sample` — done in f4a97b2
- [x] Add `BASE_URL_CART` to `services/report/.env.sample` and `services/journal/.env.sample` — done in f4a97b2
- [x] Add `TOKEN_URL: http://account:8000/api/v1/accounts/token` to the six non-account services in `docker-compose.prod.yaml` — done in f4a97b2
- [x] Add `services/stock/.env.sample` — done in f4a97b2

### 2. Align on direct HTTP and remove the duplication

- [x] Convert `BASE_URL_*` and `TOKEN_URL` in `services/docker-compose.yaml` from invoke URLs to `http://<service>:8000/api/v1` — done in abe3605
- [x] Convert `services/*/.env.sample` the same way, dropping the `${BASE_URL_DAPR}/invoke/...` form for inter-service calls while keeping `BASE_URL_DAPR` itself — done in 739b328
- [x] With the base file carrying the correct values, `docker-compose.override.yaml`'s `BASE_URL_*` block becomes pure duplication. Move its remaining content (`MONGODB_URI`, `DB_NAME_PREFIX`, `depends_on: mongodb`, `USE_GRPC` / `GRPC_*` / `MASTER_DATA_GRPC_URL`, `ports: 50051`) into the base file and delete the override — done in abe3605
- [x] Drop `COMPOSE_OVERRIDE` from `scripts/start.sh:18,85` and `scripts/stop.sh:18,78` — done in abe3605
- [x] Verify `docker compose config` is env-equivalent to the previous base+override merge apart from the intended `BASE_URL_*` values — done in abe3605

`docker-compose.prod.yaml` stays a separate file — it is the performance-measurement baseline for cart / terminal / master-data and carries genuinely prod-specific settings (multi-stage images from `Dockerfile.prod`, per-service `UVICORN_WORKERS`, `REQUEST_LOG_*` tuning, `restart: always`). Consolidating it into the base file was considered and rejected in #158.

### 3. Make the defect class non-recurring

Six occurrences across four sources is the actual problem; fixing the instances without a guard just resets the clock.

- [x] Have each service declare the `BASE_URL_*` it requires, and assert at startup that none of them is still at its `settings_web.py` / `settings_auth.py` default. Fail loudly rather than deferring to a connection error at first use. — done in 739b328
- [x] **Done in `6f4c8a7`** — Add a test or CI step asserting that the declared, supplied, and actually-used sets agree. Two directions are needed:

  1. **code → declaration** — walk each service's `app/` for its real call sites and assert they match the `REQUIRED_SERVICE_URLS` list in its `app/main.py`. The startup verification added in `739b328` only checks what a service *declares*; adding a new `get_service_client("...")` without updating the declaration passes it silently.
  2. **declaration → configuration** — assert `docker-compose.yaml`, `docker-compose.prod.yaml`, and each `.env.sample` supply that set.

  The check must recognise both `get_service_client("x")` and `get_service_client(service_name="x")`. A grep matching only the positional form under-reports: it is how `report`→cart and `journal`→cart were missed initially, which put a wrong requirement table in this issue until it was corrected.

#### `_get_service_url` swallows unknown service names

`services/commons/src/kugel_common/utils/http_client_helper.py:352-371` resolves a service name to a URL by scanning `dir(settings)`:

```python
    for url_setting in base_url_settings:
        if service_name in url_setting:
            return getattr(settings, url_setting)
    return None
```

Two problems, both feeding the same defect class:

- **An unknown service name returns `None` silently.** `HttpClientHelper.__init__` then turns it into `""` (`base_url.rstrip('/') if base_url else ""`), `_build_url` yields a relative URL, and httpx raises `UnsupportedProtocol: Request URL is missing an 'http://' or 'https://' protocol.` That is a subclass of `httpx.RequestError`, so `_make_request:146` catches it and **retries three times with a one-second delay** before surfacing a generic "Request error". A typo'd or unconfigured service name therefore fails three seconds late with no indication of the real cause.
- **Matching is by substring, not equality.** Resolution order is `dir(settings)`, i.e. alphabetical. No current pair of `BASE_URL_*` names is a substring of another, so nothing is broken today, but the behaviour is not a property worth depending on.

- [x] Make `_get_service_url` raise on an unknown service name instead of returning `None` — done in 8694d34
- [x] Resolve by exact `BASE_URL_{NAME}` match rather than substring containment — done in 8694d34
- [x] Fix the misleading docstring example at `:342`, which uses a non-existent `"inventory"` service — copying it produces exactly the silent failure above. The `Args` line two lines up already uses real names (`"cart"`, `"account"`). — done in 8694d34
- [x] Delete the `if __name__ == "__main__":` demo block at `:375-478` (104 lines). It is never imported or executed, its endpoints (`/api/inventory/status`, `/api/inventory/products/{id}`) match no real route — stock mounts everything under `/api/v1` (`services/stock/app/main.py:65-66`) — and it is the source of most of the file's `"inventory"` references. For the record, `"inventory"` is a fictional example, not a stale name for the stock service: both files have existed since the initial commit `f31efa1` and no rename ever occurred. — done in 8694d34

## Out of scope

- Dapr pub/sub, state store, and binding usage — unchanged.
- The gRPC transport for cart → master-data item lookup — unchanged, and unaffected by this issue.
- Lower-priority findings recorded in the #158 closing comment: prod sidecars passing no `-config`, components declaring no `scopes`, and `daprio/daprd` pinned to `latest`.

  **No longer applicable:** that comment also flagged the `http://localhost:3500/.*` respx catch-all in `services/cart/tests/integration/conftest.py` as able to swallow real service calls, because invoke URLs live under the sidecar port. Converting `BASE_URL_*` to direct HTTP dissolved that containment — only Dapr state and pub/sub traffic reaches port 3500 now. The catch-all still answers unmocked Dapr calls with 204 rather than failing loudly, which is a much smaller concern and is left alone.

## Verification

- [x] `./scripts/run_unit_tests.sh`, `./scripts/run_integration_tests.sh`, `./scripts/run_e2e_tests.sh` all green — verified
- [x] Report generation exercised end to end — item / category / payment makers reach master-data — verified
- [x] `grep -r 'invoke' services/docker-compose*.yaml services/*/.env.sample` returns nothing for inter-service calls — verified
- [x] Starting the stack with a service's `BASE_URL_*` deliberately removed fails at startup with a named variable, not at first request — verified
- [x] `get_service_client("does-not-exist")` raises immediately and names the unresolved service, rather than retrying three times into a generic request error — verified

kugel-masa:
Closes #159.

## What was wrong

Four independent copies of the same inter-service wiring existed and disagreed:
`docker-compose.yaml` and `services/*/.env.sample` used Dapr invoke URLs, while
`docker-compose.override.yaml` and `docker-compose.prod.yaml` used direct HTTP.
The override was applied by default, so it silently replaced the base file's
invoke URLs on every local run.

Duplicating the wiring four ways had produced the same defect six times: a
`BASE_URL_*` a service actually uses was simply absent, so it fell back to the
`settings_web.py` default (`http://localhost:800X/api/v1`), which resolves to
nothing inside a container. The resulting failure surfaced three retries later as
a generic transport error — and for `BASE_URL_TERMINAL`, as HTTP 401 "Invalid API
key" plus a false `Invalid api_key attempt` audit-log entry, because
`kugel_common.security` maps `HttpClientError` with `status_code None` onto an
auth rejection.

The transport direction was settled in #158: local (in-store / edge) deployments
use direct HTTP and gRPC. Dapr must not sit on the path of a sale — a daprd
outage cannot be allowed to stop front-of-store operations. Dapr remains in use
for pub/sub, state stores and bindings, which are asynchronous and guarded by the
circuit breaker in `DaprClientHelper`.

## Commits

| Commit | Change |
|---|---|
| `f4a97b2` | Fix the six omissions across both compose files and the `.env.sample` files |
| `8694d34` | `_get_service_url` raises on an unknown service name and matches exactly; delete a 104-line `__main__` demo block |
| `739b328` | Verify required service URLs at startup; point `.env.sample` at localhost; update both READMEs |
| `abe3605` | Fold `docker-compose.override.yaml` into the base file and delete it |
| `1b22dc1` | Integration conftests resolve respx bases from `settings` and declare their required set |
| `6f4c8a7` | Contract test across code, declaration and configuration; correct `stock`'s over-declaration |

## How recurrence is prevented

`verify_service_urls` runs in each service's lifespan, ahead of `startup_event`,
and refuses to start while naming the missing settings. Detection uses
pydantic-settings' `model_fields_set` — "was this explicitly provided" — rather
than comparing against the field default, because a deployment may legitimately
set a value equal to the default and that must not be reported as missing.

That only checks what a service *declares*, so a new test asserts the declaration
equals the actual call sites, and that both compose profiles and every
`.env.sample` supply it. It uses `ast` rather than grep for two reasons found the
hard way: both `get_service_client("cart")` and
`get_service_client(service_name="cart")` occur, and matching only the positional
form under-reports — that is how `report`→cart and `journal`→cart were missed when
#159 was first written; and example calls appear inside docstrings, which `ast`
cannot mistake for real ones. Requirements inherited through commons are
attributed to the importer, since no service's own `app/` mentions
`BASE_URL_TERMINAL` even though every API-key-authenticating service needs it.

The test found an over-declaration on its first run: `stock` listed
`BASE_URL_MASTER_DATA` but never resolves it.

## Consequence for host-run services

The bare-local workflow in the READMEs (infrastructure in docker, service under
uvicorn on the host) relied on the `settings_web.py` defaults, which are correct
for that topology. They are no longer sufficient, so `cp .env.sample .env` is now
a required step and both READMEs document it with the reason. The `.env.sample`
files were converted to `http://localhost:800X/api/v1`, matching the port mapping
they already documented. `BASE_URL_DAPR` keeps its per-service port — Dapr is
still used for pub/sub and state.

## Verification

- 2345 unit tests, the integration tier across all seven services, and all eight
  e2e suites including the cross-service scenarios — green
- Equivalence of the compose merge was checked by rendering `docker compose config`
  for the previous base+override pair and for the new base file alone: of 717
  lines the only difference is the six intended `TOKEN_URL` values, which the
  override never overrode and which therefore still pointed at an invoke URL
- Images were rebuilt and the stack restarted from the merged base file alone;
  all seven services report healthy and the six with declared requirements log
  `service URL settings verified` from inside their containers
- `cart`'s lifespan was booted from a copied `.env.sample`, and refused to boot
  without one
- `grep -r invoke services/docker-compose*.yaml services/*/.env.sample` returns
  nothing, and no reference to `docker-compose.override.yaml` remains anywhere

## Not included

Lower-priority findings recorded in the #158 closing comment remain open: prod
sidecars passing no `-config`, Dapr components declaring no `scopes`, and
`daprio/daprd` pinned to `latest`. The unauthenticated cross-tenant gRPC surface
in `REVIEW_FINDINGS.md` is a separate concern and needs its own issue.


kugel-masa (commented):
## Self-review

Reviewed the diff against a running stack rather than by reading alone. One finding was significant enough to invalidate this PR's central claim; it is fixed in `067bc99`, on this branch.

### 🔴 `account` still had the defect this PR claims to close — fixed in `067bc99`

`account` calls no other service from its own routes, so it was never wired with `REQUIRED_SERVICE_URLS` and neither compose file set `BASE_URL_TERMINAL` for it. But it installs the request-logging middleware, and `log_requests.py:30` imports `security.get_terminal_info`, which calls `get_pooled_client("terminal")`. That fires whenever a request carries an `X-API-KEY` header together with a `terminal_id` — and `terminal_id` is read from the query string (`log_requests.py:216-226`), so it is client-controlled and reachable on every route.

The lookup therefore resolved the `settings_web.py` localhost default:

```
POST /api/v1/accounts/token                                   -> 200 in 0.19s
POST /api/v1/accounts/token?terminal_id=T5837-5678-9          -> 500 in 3.22s
     with X-API-KEY: bogus-key

[ERROR] Retry failure (3 attempts): http://localhost:8001/api/v1/terminals/T5837-5678-9
[ERROR] Failed to get terminal info: Connection error
[WARNING] audit: Invalid api_key attempt via terminal service (terminal_id=T5837-5678-9)
```

A successful authentication answered 500, three seconds late, with a false audit entry — precisely the failure mode this PR's description uses to motivate the work. Not a regression from this branch, since `account` never had the setting, but squarely inside what the branch claims to have closed.

Fixed by declaring `REQUIRED_SERVICE_URLS = ["BASE_URL_TERMINAL"]` on `account` and setting the value in both compose files and `account/.env.sample`. The same request now completes in 0.24s with no retries and no audit entry.

### 🟡 The contract test had a blind spot that let it through — fixed in `067bc99`

`test_service_url_contract.py` attributed commons-inherited requirements only one import hop deep. `account` imports the middleware, not `security`, so nothing was inherited and the test passed. It now walks the import graph to a fixpoint — fixpoint rather than recursion, because the graph may contain cycles.

The test was changed first and confirmed to fail on `account` before the fix was applied. Without that ordering there is no evidence it detects anything.

### 🟡 A larger defect underneath — filed as #161

After the `account` fix the reproduction still answers 500, just quickly. The key really is invalid, so the middleware raises its own `HTTPException(401)` from the `finally` block, which escapes and is rendered as a 500.

This has nothing to do with configuration and affects every service:

| request | result |
|---|---|
| `cart`, bad key, no `terminal_id` | **401** — correct |
| `cart`, bad key, with `terminal_id` | **500** |
| `terminal`, same | **500** |
| `account`, same | **500** |

An ordinary authentication failure is reported as an internal server error whenever the request carries a `terminal_id`, and because the exception aborts the rest of the `finally` block, no request log is written for it either. `_get_current_user` on the next line is already hardened against exactly this and its docstring explains why; `_get_terminal_info` runs first and is not. Filed as #161 — out of scope here.

### ✅ Held up under review

- Compose-merge equivalence is demonstrable, not asserted: rendering `docker compose config` for the old base+override pair and the new base file alone differs in exactly the six intended `TOKEN_URL` values, out of 717 lines.
- The contract test is non-circular — declarations are `ast.Constant`, usage is `Call`/`Attribute` — which is why it could catch `stock` over-declaring `BASE_URL_MASTER_DATA` it never resolves.
- `model_fields_set` over comparing against defaults is the right call: a deployment may legitimately set a value equal to the default.
- `ServiceUrlNotConfiguredError` deliberately not subclassing `HttpClientError`, so `security`'s 401 mapping cannot swallow a misconfiguration, with a test pinning the relationship.
- No duplicate environment keys or `depends_on` blocks in either compose file after the merge.

### Left alone

- `.env.sample` still carries `# e.g. http://terminal.myserver.com/api/v1` above values now in the same form. Redundant, cosmetic.
- The contract test lives in commons but reads sibling services' files. Documented in its module docstring; a repo-root location would be cleaner if a runner is ever added for one.

### Verification after the fix

2345 unit tests, the integration tier across all seven services, and all eight e2e suites including the cross-service scenarios — green. `account` restarted from a rebuilt image and logs `service URL settings verified: BASE_URL_TERMINAL`.

Issue #165, opened 2026-08-20
cart: the signed snapshot has no freshness — an older validly-signed envelope is accepted (add a monotonic revision for rollback detection)


kugel-masa:
Found while porting #156/#157 downstream.

## Problem

`verify_envelope` (`services/cart/app/services/snapshot_service.py`, L269) checks six things:

1. the envelope is an object
2. required fields are present (L290)
3. `schema_version` is supported
4. `kid` is known
5. the signature matches
6. the cart status is restorable / not terminal

It never checks **whether the envelope is the latest one for that cart**.

`issued_at` is in the envelope and is covered by the signature, but its only consumer is `extract_audit_meta` — it is written to `cart_restore_log` and never compared against anything.

The signature proves *"we issued this, unmodified"*. It does not prove *"this is current"*. And in the stateless path the server deliberately does not consult the cache, so it has no other basis for telling a current envelope from an earlier one for the same cart.

## What is already covered

To be clear about the boundary — the finalize side is solid:

- tranlog carries a unique partial index on `(tenant_id, store_code, cart_id)`
- `tranlog_repository` returns the existing tranlog for an identical repeat finalize, and raises `FinalizeConflictException` when a *different* operation reuses a known `cart_id`

So resurrecting a pre-finalize envelope cannot produce a second transaction (#152 is done).

The gap is **rollback before finalize**. A successful carried request writes nothing server-side — `prepare_stateless_from_snapshot` only writes an audit record on rejection — so revisions 8, 9, 10 leave no trace. Finalizing the revision-7 envelope is indistinguishable from a normal finalize.

## Why this is scoped to detection, not prevention

Rejecting a stale envelope synchronously would require the server to know the high-water state for that cart, i.e. a per-request write. That gives back what phase 2 was built to remove, so it is deliberately out of scope here.

This issue proposes making rollback **reliably visible after the fact**, at no additional write.

## Proposal

1. **Carry a monotonic `revision: int` in the envelope** — incremented once per cart mutation, covered by the signature, `schema_version` bumped to 2.
2. **Record it as a dedicated field on `RequestLog`** — not inside the logged body, which #155 proposes to strip.
3. Rollback then shows up as a non-monotonic `revision` sequence for a `cart_id` (…8, 9, 10, 7) and is findable by query. Normal operation is strictly increasing; a repeat is an equal value.

Cost: one signed integer in the envelope, one indexed field on a record that is already written for every request.

[...]

PR #167, opened 2026-08-20
fix(commons): bound logged request/response bodies (#155)

### LINKED ISSUE #155: commons: request-log middleware stores full response bodies — snapshot attachment roughly doubles log volume
Found during the second-pass review of #154 (signed cart snapshots, #148).

## Problem

`kugel_common/middleware/log_requests.py` captures the **entire response body** for every request and persists it twice:

- to the request log **file** (`_output_request_log_to_file`, logs `response_info.body` verbatim), and
- to the per-tenant MongoDB **`request_log` collection** (buffered `insert_many`).

There is no truncation or field filtering on the response side (`_make_response_info` stores the parsed body as-is).

With #148 merged, every cart-mutating response additionally carries `signedSnapshot` — the full cart document **including embedded masters** — so the logged body roughly doubles (desk estimate: ~40–80 KB raw for a 40-item cart, before the cart payload itself). That growth lands on every mutating request, across both the log file and Mongo, and is invisible to the gzip transport win from #147 (logs store the uncompressed parsed body).

## Impact

- `request_log` collection and log file growth scale with snapshot size × mutating request rate.
- Log noise: the same cart document is stored twice per response (response fields + snapshot copy).

## Mitigation options (shared middleware, affects all services)

1. **Strip known-bulky fields** from the logged response body (`signedSnapshot` / `signed_snapshot`) — cheapest, targeted, keeps the rest of the body intact for debugging.
2. **Truncate** logged bodies at a configurable byte budget (e.g. 16–64 KB) with a `truncated: true` marker.
3. Both: strip the snapshot and keep a small truncation backstop.

Option 1 + 3's backstop seems right: the snapshot is reconstructible (it is what the client received) and has no forensic value beyond the response fields already logged; the audit trail for restores lives in `log_cart_restore`.

## Sizing

Should be measured together with **T029** of #148 (`specs/148-cart-snapshot-restore/tasks.md`, pending since #147 is now merged): the same perf run that validates SC-005/SC-006 can capture per-request `request_log` document sizes before/after mitigation.

## References

- #148 / PR #154 (snapshot attachment), second-pass review comment
- #147 / #149 (gzip — transport only, does not affect stored log size)

kugel-masa:
Closes #155.

## Problem

`kugel_common/middleware/log_requests.py` stored every request and response body verbatim, twice — in the request log file and in the per-tenant `log_request` collection. Since #148/#156 every cart-mutating response carries `signedSnapshot` (the full cart document, embedded masters included), so the logged volume roughly doubled on the response side. gzip (#147) never bounded it: the log stores the parsed, uncompressed body.

## What this does

Both the request and the response body are sanitized before they are stored:

1. **Field stripping** — fields named in `REQUEST_LOG_STRIP_FIELDS` (default `signedSnapshot,signed_snapshot`) become a marker that keeps only the value's *short scalar* members. `kid` / `schema_version` / `issued_at` — and a future monotonic `revision` (#165) — stay queryable; `cart_document`, `signature` and long strings do not. The envelope is reconstructible and the restore audit trail lives in `log_cart_restore`.
2. **Size backstop** — a body still above `REQUEST_LOG_MAX_BODY_BYTES` (default 32 KB) is stored as `{"_truncated": true, "_encoded_bytes": N, "_preview": "…"}`, with the preview clamped to the budget so the setting is a real ceiling.

Sanitization never raises and never affects what the client receives.

| Variable | Default | Purpose |
|---|---|---|
| `REQUEST_LOG_STRIP_FIELDS` | `signedSnapshot,signed_snapshot` | Comma-separated fields replaced by a marker; empty disables |
| `REQUEST_LOG_MAX_BODY_BYTES` | `32768` | Ceiling for a logged body; `0` disables |

## Performance

The walk runs on every request of every service, so the raw bytes are threaded through to skip work whose outcome they already determine: a byte-level scan avoids walking a body that mentions no stripped field, and an untouched over-budget body is measured from the bytes it already has.

| 282 KB response | before | after |
|---|---|---|
| sanitize cost | 3.51 ms | **0.12 ms** |

(For reference, `json.loads` of the same body is 1.14 ms.)

## Measured effect (full e2e run, live stack)

Cart `log_request` documents carrying a response snapshot:

| | count | avg | max |
|---|---|---|---|
| before | 687 | 5,758 B | 14,482 B |
| after | 500 | **2,460 B** | **6,547 B** |

500 of 500 stripped, 0 unstripped. Of **6,898** logged requests across all services, only **3** responses hit the 32 KB backstop — all transaction-list reads (189/66/48 KB); no ordinary POS operation was truncated, and no request body was.

## Note on the request side

Cart was already covered: `SnapshotEnvelopePeelMiddleware` (#157) runs outside the logger, and 0 of 2,134 mutating requests in the e2e run carried a snapshot into the log. The strip still applies there for paths the peel does not cover, and for the other six services.

## Behaviour change to be aware of

Responses larger than 32 KB (in practice: transaction-list reads) are no longer logged in full. Set `REQUEST_LOG_MAX_BODY_BYTES=0` to restore the previous behaviour.

## Testing

- **50 new unit tests** — sanitizer semantics, middleware wiring for *both* sinks (file + Mongo buffer), budget behaviour, depth cap, non-JSON and body-less requests, multibyte previews
- unit: 8/8 service suites pass (commons 499)
- integration: 7/7 suites pass
- e2e: 8/8 suites pass (cart 59, cross-service 31)
- `ruff check` / `ruff format` clean on new and changed files

kugel_common bumped to **0.1.66**; every service Pipfile/lock updated.


kugel-masa:
## Code Review (PR #167)

Second pass over the branch diff, with the runtime behaviour checked against a real `Settings` instance and against the `log_request` documents produced by the full e2e run.

### Findings

#### 🟡 Warning — fixed in 0a10d67

- **[Docs/Usability]** `settings_app.py`, commons spec (en/ja) — the documented off-switch did not work. `REQUEST_LOG_STRIP_FIELDS=""` **does not** disable stripping: `kugel_common`'s `Settings` runs with `env_ignore_empty=True`, so an empty value is dropped and the default applies. Verified:

  ```
  REQUEST_LOG_STRIP_FIELDS=""  -> 'signedSnapshot,signed_snapshot'   # ignored, default wins
  REQUEST_LOG_STRIP_FIELDS=" " -> ()                                  # actually disables
  ```

  Documented the working value in the settings docstring and both spec translations. A word like `off` would also "work" but would make `_may_contain` scan for a substring that occurs in ordinary payloads, so it would silently cost the walk it is meant to skip — the space is the right answer.

- **[Test quality]** `test_log_body_sanitizer.py` — the depth-cap test asserted `_truncated is True` with a 4 KB budget, which fires whether or not the cap left the snapshot in place, so it did not cover the documented limit at all. Split into two: one pins the cap's actual behaviour (a snapshot nested deeper than 32 keeps its `cart_document`), one pins the backstop that bounds it.

#### 🔵 Suggestion — not changed

- **[Forensics]** `log_utils.py` — the marker drops `signature`. That is right for volume and the issue calls it out, but it does mean the request log no longer records which signature the server issued for a given response. The server can re-sign, and rejections are audited in `log_cart_restore`, so this is a conscious trade-off rather than a gap.
- **[Types]** `parse_log_strip_fields() -> tuple` and `_strip_fields(fields: frozenset, ...)` could be `tuple[str, ...]` / `frozenset[str]`. Matches the surrounding file's looser style as written.
- **[Robustness]** the marker bounds the *length* of retained scalars (128 chars) but not their *count*; with `REQUEST_LOG_MAX_BODY_BYTES=0` there is then no bound at all. No real payload looks like that today.
- **[CI]** this repository reports no checks on PRs, so the test evidence below is local only.

### Verified during this pass

- **The off-switch and the budget knob take effect from the environment** — `REQUEST_LOG_STRIP_FIELDS=foo,bar` / `REQUEST_LOG_MAX_BODY_BYTES=4096` are picked up by a fresh `Settings()`.
- **Nothing reads request-log bodies back** — the `log_request` collection is written by the middleware and created at tenant setup; no service queries it. Stripping cannot break a consumer.
- **Both sinks carry the marker** — confirmed in Mongo (500/500 recent snapshot-carrying responses stripped, 0 unstripped) and in the container's `/app/request.log` (260 marker lines, no `cart_document`).
- **The 32 KB default does not over-trigger** — of 6,898 logged requests across all services in the e2e run, 3 responses were truncated (transaction-list reads, 189/66/48 KB) and 0 request bodies.
- **Side benefit**: a body can no longer grow a `log_request` document toward the 16 MB BSON limit.

### Testing

| tier | result |
|---|---|
| unit | 8/8 suites pass (commons 500, incl. 51 new) |
| e2e | 8/8 suites pass (cart 59, cross-service 31) |
| ruff | check + format clean on new and changed files |

Not covered: latency under load (the 3.51 ms -> 0.12 ms figure is a microbenchmark, not a `/perf-test` comparison against `main`).

### Verdict

**Comment only** — the two 🟡 findings are fixed in this branch; nothing blocking remains. I authored the change, so this is not an independent approval.

Issue #168, opened 2026-08-20
cart: DUAL mode has two independent receipt-number sources — a degraded snapshot silently reissues numbers


kugel-masa:
Split out of #166 (see the design comment there).

## Problem

The carried path and the legacy path draw receipt numbers from **different counters**:

| path | source | durable home |
|---|---|---|
| carried (phase 2) | the terminal's own counter, seeded at open | terminal service, `terminal.receipt_no` |
| legacy (phase 1 / no snapshot) | `numbering_count(CounterType.Receipt, start, end)` (`cart/app/services/tran_service.py:203`) | cart's `terminal_counter` collection |

Nothing reconciles the two, and the branch is taken **per transaction**, not per terminal:

```python
carried = cart.transaction_datetime is not None   # tran_service.py:186
```

## Why a phase-2 terminal can fall into the legacy path

Snapshot signing degrades silently by design:

> `SNAPSHOT_HMAC_KEYS` が未設定の場合、機能は縮退する: スナップショットは発行されず（レスポンスの `signedSnapshot` が null）、提示されたエンベロープは全て拒否される
> — `docs/ja/cart-snapshot-key-rotation.md`

A key that is unset, malformed, or mid-rotation therefore stops snapshots from being issued. The client has nothing to carry, sends a legacy request, and cart numbers it from the *other* series — which starts at `RECEIPT_NO_START_VALUE` and knows nothing about how far the terminal's own series has advanced.

## Impact

Numbers already printed on customer receipts can be **issued a second time**. That is worse than the gaps #166 discusses, which are all abnormal-or-deliberate and detectable; this one produces two transactions carrying the same receipt number with no marker distinguishing them. `receipt_no` is not index-enforced anywhere (`journal/app/database/database_setup.py:117` is `unique: False`), so nothing rejects it.

It is also invisible in the happy path: the only warning today is a startup log line about the key ring, not a per-transaction signal.

## Options

1. Point the legacy path at the same durable home, so both series advance one counter.
2. Reject (or loudly flag) a mutating request from a terminal that has already used the carried path, instead of silently falling back — i.e. make the fallback per *terminal*, not per transaction.
3. Emit a per-transaction warning when a finalize takes the legacy path while snapshot signing is degraded, and surface it in the audit trail.

(1) is the durable fix; (3) is the cheap detection that makes the incident visible in the meantime.

## References

- #166 — receipt-number range and rollover on the carried path (design comment there)
[...]

PR #169, opened 2026-08-20
feat(cart,terminal): derive receipt numbers from a running counter (#166)

### LINKED ISSUE #166: cart,terminal: the carried numbering path ignores the configured receipt-number range, and max-reconcile at open cannot survive a rollover
Found while porting #156/#157 downstream.

## Problem

The receipt-number range is an existing, configurable feature:

```python
# kugel_common/config/settings_app.py
RECEIPT_NO_START_VALUE: int = 111111
RECEIPT_NO_END_VALUE:   int = 999999   # cycles back to start
```

It is overridable per tenant through the master-data settings hierarchy, and the central numbering path implements it properly — `TerminalCounterRepository.numbering_count` does document creation, field initialisation, increment **and rollover at `end_value`** in a single atomic aggregation pipeline.

Phase 2's carried path bypasses all of it. In `tran_service.create_tranlog_async`:

```python
carried = cart.transaction_datetime is not None
if carried:
    tranlog.receipt_no = cart.receipt_no          # whatever the terminal carried
else:
    receipt_start_raw = await self._get_setting_value_async("RECEIPT_NO_START_VALUE")
    receipt_end_raw   = await self._get_setting_value_async("RECEIPT_NO_END_VALUE")
    tranlog.receipt_no = await self.terminal_counter_repository.numbering_count(
        countType=CounterType.Receipt.value, start_value=..., end_value=...)
```

and `open_terminal_async` seeds the terminal-owned counter with a bare zero:

```python
if getattr(terminal, "receipt_no", None) is None:
    terminal.receipt_no = 0
```

The code comment acknowledges this ("A configurable start value / rollover is a follow-up") but there is no issue tracking it.

## Impact

With default settings, receipts today are six digits starting at 111111 and cycling at 999999. On the carried path they are 1, 2, 3, … — a customer-visible change, and a configured setting that silently stops taking effect. DUAL mode hides it while legacy clients still exist; moving to `CART_REQUEST_SNAPSHOT_MODE=REQUIRED` makes it universal.

## This is not a straight port — rollover and max-reconcile conflict

The open-time reconcile is a plain `max()`:

```python
terminal.receipt_no = max(getattr(terminal, "receipt_no", None) or 0, client_receipt_no)
```

A terminal that reaches `end_value` and rolls back to `start_value` then opens again:

```
stored = 999999, carried = 111111
max(999999, 111111) = 999999      # the rollover is undone
```

The cycle breaks on its first wrap, permanently. And the guard only looks upward:

```python
if carried > stored + MAX_COUNTER_JUMP:   # a carried value far BELOW stored is not checked
```

so nothing warns — the number silently jumps back to the top of the range.

So implementing "the same thing on the new path" is not sufficient. The durable home needs a representation that survives a wrap — for example storing the in-range position together with a cycle count and reconciling on `(cycle, position)`, or scoping the `max()` to within a cycle. That is a design decision, which is why this is filed rather than left as a code comment.

**`business_counter` is not affected**: it is monotonic and never resets, so `max()`-reconcile is correct for it. The problem is specific to `receipt_no`, which is the one carried counter that is *meant* to wrap.

## References

- #146 — client-carried cart design umbrella
- #156 / #157 — phase 2, where the carried numbering path was introduced
- `kugel_common/config/settings_app.py` (range defaults)
- `cart/app/models/repositories/terminal_counter_repository.py` — `numbering_count` (the working implementation)
- `cart/app/services/tran_service.py` — the `carried` / `else` split
- `terminal/app/services/terminal_service.py` — `open_terminal_async`, `__check_carried_counter_jump`

kugel-masa:
Closes #166. Design decision and reasoning: [#166 comment](https://github.com/kugel-masa/kugelpos-backend/issues/166#issuecomment-5352415562).

## Problem

Phase 2 moved receipt numbering to the terminal and the configured range went with it. `RECEIPT_NO_START_VALUE` / `RECEIPT_NO_END_VALUE` stopped taking effect on the carried path, which recorded whatever number the client sent — 1, 2, 3… instead of six digits cycling at 999999.

Porting the server-side rollover across would not have been enough. The open-time reconcile is a `max()`, and a number that wraps cannot be compared that way:

```
stored = 999999, carried = 111111  →  max() = 999999   # the wrap is undone, permanently
```

and the guard only looked upward (`carried > stored + MAX_COUNTER_JUMP`), so nothing warned.

## What this does

The terminal carries **one running counter** instead of the printed number:

```
receipt_counter = receipt_base(epoch) + seq
receipt_no      = start + ((receipt_counter - 1) mod (end - start + 1))
```

Keeping the wrap in the *derivation* is what makes the rest work. A running count only increases, so **the existing `max()` reconcile and `MAX_COUNTER_JUMP` guard stay correct and unchanged**, and today's stored values migrate by being read — clients counted 1, 2, 3 with no wrap, which is already a running counter.

**The transaction key is untouched**: `(business_counter, transaction_no)` as shipped, no downstream index rework.

| layer | change |
|---|---|
| commons | `kugel_common.utils.receipt_numbering` (`derive_receipt_no` / `receipt_cycle`), `receipt_counter` on `BaseTransaction` and `TerminalInfoDocument`, `receipt_counter` claim in the terminal token |
| terminal | open accepts `receipt_counter`; the pre-#166 `receipt_no` is the same value and is still accepted; both fields kept in step |
| cart | `FinalizeContext` and the finalize-context envelope carry it; the carried path derives the printed number from it and the configured range; the transaction log records it behind a **non-unique** index |

A client that also sends the number it printed keeps that number — it is on paper — and a disagreement is logged rather than silently corrected. The range itself needs no new delivery mechanism: a client reads it from master-data like any other terminal setting.

## Naming

`receipt_counter`, deliberately **not** `transaction_counter`. The value is client-owned, gappy and not unique (spec 156 Q58 settles for audit detection over enforcement), and a `transaction_*` name would invite the keying mistake spec 156 already had to correct ("`transaction_no` 単独キーにしている箇所は是正対象"). The field comment states what it does not guarantee.

Gaps have exactly three causes — a safe jump at terminal replacement, an offline finalize that never arrived, or a reconcile where the stored value was higher. Nothing in normal operation produces one, so a hole in the counter is a signal worth investigating. (A cancellation is a finalize like any other and consumes a number; abandoning a cart consumes nothing.)

## Verified against the running stack

```
counter=1       →  receipt_no=111111
counter=5       →  receipt_no=111115
counter=888889  →  receipt_no=999999     # last number of the cycle
counter=888890  →  receipt_no=111111     # wrapped, same terminal
```

Both 111111 rows exist on one terminal and are told apart by the counter — which is exactly why `max(receipt_no)` cannot reseed a replacement terminal and `max(receipt_counter)` can.

## Compatibility

Pre-#166 clients are unaffected: they carry no counter, send their value under the old `receipt_no` name at open, and their receipt numbers are recorded as sent. Covered by tests on both sides.

## Testing

45 new tests — commons arithmetic (wrap, cycle, boundaries), terminal reconcile (wrap survival, stale value, jump guard, legacy field), cart derivation (range from settings, disagreement logging, misconfigured range), and e2e through the live stack (counter → printed number, including the wrap).

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |

kugel_common 0.1.68; every service Pipfile/lock updated.

## Out of scope

#168 — DUAL mode draws receipt numbers from two independent counters, so a degraded snapshot key silently reissues numbers from the other series. Worse than a gap, and independent of this fix.


kugel-masa:
## Code Review (PR #169)

Reviewed against the running stack rather than by reading alone. Two things the implementation got wrong, both fixed in e1c7aa5, plus one defect this work uncovered that belongs elsewhere.

### Findings

#### 🔴 The invariant this PR documents does not hold — fixed in e1c7aa5

The docs and the `receipt_counter` field comment claim `1 tranlog = 1 seq = 1 receipt_counter = 1 printed receipt_no`, with cancellation called out as "a finalize like any other". It is not.

`POST /carts/{cart_id}/cancel` takes no finalize context (`cart.py:137` calls `cancel_transaction_async()` with no arguments), and `create_tranlog_async` branches on `carried = cart.transaction_datetime is not None` — stamped by the client at **bill**. A cancellation happens before that, so it is always numbered from the server-side series.

Confirmed in live data (33 cancelled tranlogs, all with `receipt_counter: null`), and reproduced directly:

| transaction | business_counter | transaction_no | receipt_no | receipt_counter |
|---|---|---|---|---|
| carried sale (`seq=1`) | 1 | 1 | 111111 | 1 |
| cancellation, same epoch | 1 | 1 | 111111 | null |

The docs, the gap list, and the field comment now name cancellation as the exclusion.

#### 🔴 An unavailable range printed a number outside it, silently — fixed in e1c7aa5

`_get_setting_value_async` catches its own failure and returns None, so a cache miss while master-data is unreachable left `_receipt_range_async` returning `(1, sys.maxsize)`. The carried path then derived the raw counter: counter 888890 would print **888890** instead of 111111, on the finalize path, with nothing in the log.

`_receipt_range_async` now returns whether the range is the configured one. When it is not, the number the terminal printed wins (which is why a client should send both) and the fallback is logged at ERROR. A client that sends only a counter still gets the raw value — and now the log says so.

This matters more than it looks: the carried path exists to keep working when the backend does not, so its degraded behaviour needs to be deliberate rather than incidental.

#### 🔵 Not changed

- `receipt_cycle` was exported and used only by tests; it now appears in the mismatch warning, so a disagreement says which cycle it belongs to.
- The carried finalize path performs two settings lookups it did not before. They go through the cached `SettingsMasterWebRepository` (`cache_namespace = "settings_master"`), so steady state is a cache hit. Pinning the range at open, alongside the counter in the token, would remove even that — worth considering if the finalize path is ever measured.

### Filed separately

**#170** — a cancelled sale can take a transaction identity a carried sale in the same open epoch already used. The unique index on `(tenant_id, store_code, terminal_no, business_counter, transaction_no)` should reject it, which means the cancellation *fails*; the run above happened to hit a tenant database whose indexes had not been created, so the duplicate simply persisted. Both outcomes are wrong. `spec.md` (156) Q44 fixed exactly this hazard for void/return and cancellation was not included. It also means #168's interleaving is not confined to the degraded-key case — on a phase-2 terminal every cancellation draws from the other series.

### Verified during this pass

- **The new index reaches existing collections.** `create_some_collection` ensures indexes on both new and existing ones, and after a full cart e2e run `db_cart_T6216.log_tran` carries `{tenant_id, store_code, terminal_no, receipt_counter}` alongside the #156 indexes.
- **Nothing else reads `terminal.receipt_no`** server-side, so repurposing it as the counter cannot mislead a consumer; the claim keeps feeding pre-#166 clients the value they expect.
- **Mixed versions are safe**: an old server verifying a new finalize-context envelope signs and checks whatever payload is presented and ignores the extra key; an old client's `receipt_no` is read as the counter it already was.
- **Live derivation**, one terminal: counter 888889 → 999999, counter 888890 → 111111. Both rows exist and are told apart by the counter — which is the argument for storing it.

One caveat for anyone debugging here: cart's e2e conftest drops the tenant database at session start, so running a **single** e2e file leaves `log_tran` without its unique indexes until the test that recreates the tenant runs. That is how the duplicate identity above persisted instead of being rejected.

### Testing after the fixes

| tier | result |
|---|---|
| unit | 8/8 suites pass (cart 598) |
| e2e | 8/8 suites pass |

kugel_common 0.1.69.

### Verdict

**Comment only** — both 🔴 findings are fixed in this branch. I authored the change, so this is not an independent approval.

kugel-masa:
### Correction to the review above

Running the flow against the real APIs showed one of my earlier claims to be wrong.

I wrote that an unavailable range "left `_receipt_range_async` returning `(1, sys.maxsize)`" and that "counter 888890 would print **888890** instead of 111111". That does not happen. `_get_setting_value_async` returns `get_setting_value(...)`, which falls back to the service's own settings module (`app/utils/settings.py:55`, `getattr(settings, name, None)`), and `RECEIPT_NO_START_VALUE` / `RECEIPT_NO_END_VALUE` are defined there via commons `AppSettings`. So a master-data miss resolves to the shipped defaults, not to an unbounded range.

Observed directly: the walkthrough tenant has no `RECEIPT_NO_START_VALUE` record at all — master-data answers **404** — and the finalize path still numbered correctly, `receiptCounter=888889 → 999999` and `888890 → 111111`.

The guard added in e1c7aa5 is therefore narrower than I described: it fires only if the name is missing from the service settings too, i.e. a rename or a deleted default, not a master-data outage. It stays, because reporting that case beats printing a raw counter, but it is a defensive backstop rather than the fix for a reachable defect. The `_encoded_bytes` rename and the preview clamp in that commit are unaffected.

PR #171, opened 2026-08-20
fix(cart): number a cancellation from the terminal's series (#170)

### LINKED ISSUE #170: cart: a cancellation is numbered from the server series and can duplicate a carried transaction identity
Found while reviewing #169 (the #166 fix).

## Problem

`POST /api/v1/carts/{cart_id}/cancel` takes **no finalize context**:

```python
cart_doc = await cart_service.cancel_transaction_async()   # cart/app/api/v1/cart.py:137
```

`cancel_transaction_async` builds the transaction log from the cart document alone, and `create_tranlog_async` branches on

```python
carried = cart.transaction_datetime is not None            # tran_service.py:186
```

`transaction_datetime` is stamped by the client at **bill**. A cancellation happens before that, so `carried` is always False and a cancelled sale is numbered from the **server-side** counters — `transaction_no` from `CounterType.Transaction` and `receipt_no` from `CounterType.Receipt` — even when the terminal is running the stateless path and carrying snapshots.

## Why it matters

The server `transaction_no` counter shares a key space with the carried per-open `seq`. `spec.md` (156) Q44 identified exactly this hazard for void/return and fixed it by carrying a signed finalize-context envelope:

> void/return は依然サーバ側 `terminal_counter`（連続・非リセット）で `transaction_no` を採番していた。これは carried sales の per-open seq（1,2,3…）と番号空間が異なり、同一開設セッション内で `(business_counter, transaction_no)` がユニークインデックス上で衝突しうる（採番空間が小さい新規端末ほど顕著）

Cancellation was not included in that fix.

## Observed

Against the running stack, one terminal, one open epoch:

| transaction | business_counter | transaction_no | receipt_no | receipt_counter |
|---|---|---|---|---|
| carried sale (`seq=1`) | 1 | 1 | 111111 | 1 |
| cancellation | 1 | 1 | 111111 | null |

Two different transactions, same identity. The unique index on `(tenant_id, store_code, terminal_no, business_counter, transaction_no)` is what should stop this: with it present the second insert is rejected and the **cancel fails**; the run above happened to be against a tenant database whose indexes had not been created, so the duplicate simply persisted. Both outcomes are wrong — a cancellation must not be refused, and it must not share an identity with a sale.

The receipt number is duplicated the same way, and `receipt_no` is not index-enforced anywhere.

## Scope note

This also means the interleaving described in #168 is not confined to the degraded-key case: on a phase-2 terminal, **every cancellation** draws from the other series in normal operation.

## Suggested fix

Give cancel the same treatment void/return got: accept a finalize context (`seq` / `receipt_counter` / `transaction_datetime`, and a stable `cart_id`) so a cancelled sale is numbered from the terminal's per-open series like every other transaction it produces. The cart already carries a snapshot on the cancel request, so the envelope path exists.

## References

- #166 / #169 — receipt counter (the invariant this breaks; documented there as an exclusion)
- #168 — DUAL mode's second numbering source
- #156 / #157 — phase 2; spec 156 Q44 fixed the same hazard for void/return

kugel-masa:
Closes #170. Stacked on #169 (it uses the `FinalizeContext.receipt_counter` added there); GitHub retargets this to `main` when #169 merges.

## Problem

A cancellation writes a transaction log and prints a receipt — it is a finalize. But `POST /carts/{cart_id}/cancel` took no finalize context, and `create_tranlog_async` branches on

```python
carried = cart.transaction_datetime is not None   # stamped by the client at bill
```

so a cancelled sale was **always** numbered from the server-side counters, whose `transaction_no` shares the `(business_counter, transaction_no)` key space with the carried per-open `seq`.

Observed on the running stack before this change — one terminal, one open epoch:

| transaction | business_counter | transaction_no | receipt_no | receipt_counter |
|---|---|---|---|---|
| carried sale (`seq=1`) | 1 | 1 | 111111 | 1 |
| cancellation | 1 | 1 | 111111 | null |

Where the unique index exists the second insert is rejected and **the cancellation fails**. Where it does not, the duplicate persists *and propagates*: the journal service's tenant setup then cannot build its own unique index —

```
Required index (tenant_id, store_code, terminal_no, business_counter, transaction_no)
missing on log_tran after migration (index ensure failed — likely existing data
violates a new unique constraint)
```

— which is how this surfaced during testing, as a 500 from journal's tenant creation.

## Fix

Cancel accepts the same `FinalizeContext` bill does and applies it on the stateless path, so `carried` is true and the transaction is numbered from the terminal's series like every other one it produces. No new transport: the cancel request already arrives as `{"signedSnapshot": …, "payload": …}` and the payload was simply empty.

- Same guard as bill — a context without a signed snapshot is rejected, because unsigned numbers are whatever the caller typed.
- A stateless cancel that carries no context still falls back to the server counters and now **logs a warning naming the collision**, rather than doing it silently.
- The legacy (snapshot-less) path is unchanged.

`spec.md` (156) Q44 fixed this same hazard for void/return; cancellation was not included. The #166 docs that named cancellation as an exclusion are updated — it is inside the invariant again.

## Client side

The terminal must advance `seq` and `receipt_counter` for a cancellation and stamp the time, exactly as at bill. Per the issue discussion this lands together with the client's stateless work, so no interim server-side allocation guard is included.

## Testing

- **5 unit** — the context lands on the cart document (which is what selects the carried branch), the tranlog is still written, the context-less stateless path warns, and the no-snapshot guard rejects.
- **3 e2e** — a carried cancel is numbered from the terminal's series, a sale and a cancellation in one open session take distinct identities, and a legacy cancel is unaffected.

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |

kugel_common 0.1.70.

### Test-suite note

The two numbering e2e modules now remove their synthetic high-`seq` rows afterwards. The transaction list defaults to `sort=transaction_no:-1` (`api/v1/tran.py:207`), so rows carrying `seq` in the 8800/9000 range occupied the whole of a `limit=10` first page and pushed another suite's transactions off it — that suite then failed for no reason of its own.


kugel-masa:
## Code Review (PR #171)

Reviewed against the running stack. Two gaps, both closed in 7eb26be; nothing in the change itself needed correcting.

### Findings

#### 🟡 A retried cancellation had no test — added in 7eb26be

The bill path pins lost-ACK idempotency (`test_retried_carried_finalize_is_idempotent`). This PR gives a cancellation the same numbering, so it inherits the same requirement, and nothing was holding it.

It does hold: `create_tranlog_async`'s `cart_id` pre-check calls `__is_same_finalize`, which compares operation identity (type + cancelled flag), so a repeat cancel matches and the existing record is returned rather than raising `FinalizeConflictException`. Verified end to end — both calls return the same `transactionNo` / `receiptNo`, and exactly one transaction log exists for the cart. Now covered by a test instead of by inspection.

#### 🟡 The generated API docs were stale — regenerated in 7eb26be

This branch changes the API surface in three places and none of it had reached the generated artefacts:

- `POST /carts/{cart_id}/cancel` gained a request body (this PR)
- `receiptCounter` on the bill finalize context (#166 / #169)
- `receiptCounter` on the terminal open request (#169)

`docs/openapi/*.json` and `docs/{en,ja}/*/api-specification.md` are generated and committed here — #157 regenerated them for the same reason. Re-exported from the running services and regenerated for cart and terminal. The terminal half belongs to #169's change; it lands here because this branch is stacked on it, so merging #169 alone leaves main briefly without it.

#### 🔵 Not changed

- **`CART_REQUEST_SNAPSHOT_MODE=REQUIRED` does not require the finalize context.** The dependency rejects a snapshot-*less* mutating request, but a snapshot-carrying finalize with no context still falls through to the server counters — for bill as much as for cancel. That is the interleaving #168 is about, so it belongs there rather than here; this PR at least makes the cancel case say so in the log.
- The e2e cleanup deletes the synthetic rows from cart only; the copies propagated to journal/report/stock stay. Those suites filter by their own data and pass, so it is not worth the extra coupling.

### Verified during this pass

- **Ordering matches bill exactly** — guard → load cart → `check_event_sequence`, so the same request produces the same error precedence on both endpoints.
- **The guard is the right one**: a finalize context without a signed snapshot is rejected on cancel as it is on bill; unsigned numbers are whatever the caller typed.
- **DUAL mode is unaffected** — a snapshot-less cancel takes the server counters exactly as before, covered by `test_legacy_cancel_is_unaffected`.
- **No collision remains** — a sale and a cancellation in one open session now take distinct identities, which is the whole point of the issue.
- `ruff check` / `format` clean on the changed regions (the pre-existing E402 in `cart_service.py` is untouched).

### Testing

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass (cart 68) |

### Verdict

**Comment only** — both findings are closed in this branch. I authored the change, so this is not an independent approval.

PR #175, opened 2026-08-20
fix(master-data): seed the receipt-number range so a client can read it (#174)

### LINKED ISSUE #174: cart,master-data: a client cannot read the effective receipt-number range — master-data 404s and the service default is invisible
Found while walking the #166 flow through the real APIs.

## Problem

Since #166 the terminal derives the printed receipt number itself:

```
receipt_no = RECEIPT_NO_START_VALUE + ((receipt_counter - 1) mod range_width)
```

so the client needs the effective range. The documented way to get a terminal setting is master-data:

```
GET /api/v1/tenants/{tenant_id}/settings/{name}/value?store_code=…&terminal_no=…
```

But `RECEIPT_NO_START_VALUE` / `RECEIPT_NO_END_VALUE` are **not registered as settings**. On a live tenant:

```
GET .../settings/RECEIPT_NO_START_VALUE/value  ->  404 "settings with name RECEIPT_NO_START_VALUE not found"
```

The server still numbers correctly, because `get_setting_value` falls back to the service's own configuration (`cart/app/utils/settings.py:55`, `getattr(settings, name, None)`), where commons `AppSettings` defines `111111` / `999999`. That fallback is invisible over the API.

So server and client resolve the range through different paths: the server has a two-step chain (master-data → service default), the client can only see step one. A client that hardcodes a different default prints numbers the backend disagrees with — the finalize path records the client's number (it is the one on paper) and logs a mismatch, so this shows up as a stream of warnings and a receipt series that does not match the configuration.

## Options

1. **Seed `RECEIPT_NO_START_VALUE` / `RECEIPT_NO_END_VALUE` into master-data at tenant creation**, so the documented lookup always answers. Smallest change, keeps one resolution path visible.
2. **Return the service default from the settings endpoint** instead of 404 when the tenant has not overridden a known setting. Bigger blast radius — it changes the contract for every setting.
3. **Deliver the range alongside the counter** in the terminal token / open response, so the client gets both from the same place it gets `receipt_counter`.
4. Document the defaults and require every deployment to seed them (weakest — a missed deployment step prints wrong numbers silently).

(1) or (3) look right; (3) has the advantage that the two values a client needs to number a receipt then arrive together.

## Impact

Blocks the client side of the stateless migration: the terminal cannot compute a printed receipt number it can justify without knowing the configured range.

## References

- #166 / #169 — where the derivation moved to the client
- `services/cart/app/utils/settings.py:55` — the invisible fallback
- `services/commons/src/kugel_common/config/settings_app.py` — the defaults

kugel-masa:
Closes #174. Found while walking the #166 flow through the real APIs.

## Problem

Since #166 the terminal derives its printed receipt number itself:

```
receipt_no = RECEIPT_NO_START_VALUE + ((receipt_counter - 1) mod range_width)
```

so it needs the effective range. The documented way to read a terminal setting is master-data — but the two values were never registered as settings, so on a live tenant:

```
GET /api/v1/tenants/T6216/settings/RECEIPT_NO_START_VALUE/value  ->  404
```

The backend still numbered correctly, because `get_setting_value` falls back to the service's own configuration (`cart/app/utils/settings.py:55`), where commons `AppSettings` defines `111111` / `999999`. **That fallback is invisible over the API.** Server and client resolved the range through different paths and the client could only see the first step — it could not compute a receipt number it could justify, and a client that guessed a different default would print a series the backend disagrees with.

## Fix

Tenant setup seeds both ends with the shipped defaults, **insert-if-absent**:

- values read from commons `AppSettings`, not restated here — this service does not inherit it, and a second copy would drift from the one the services actually number with
- re-running setup (how index migrations reach existing tenants) never overwrites an operator's own value
- a failed seed logs and continues: a tenant without its defaults still works via the service fallback, and failing setup over it would be worse

Existing tenants pick the settings up by re-running `POST /tenants`.

## One bug caught on the way

The first version inserted the documents directly into the collection. They looked right in Mongo — and then `GET /settings` started returning **422**:

```
1 validation error for BaseSettingsMasterResponse
entry_datetime  Input should be a valid string [input_value=None]
```

The response schema renders `entry_datetime` from `created_at`, which only the repository stamps. Seeding now goes through `SettingsMasterRepository`, and a test pins that.

## Verified against the running stack

```
before:  GET .../settings/RECEIPT_NO_START_VALUE/value  ->  404
POST /tenants                                          ->  201
after:   RECEIPT_NO_START_VALUE  ->  200  {"value": "111111"}
         RECEIPT_NO_END_VALUE    ->  200  {"value": "999999"}
         GET /settings           ->  200  (list intact)
```

## Testing

- **7 unit** — both ends seeded with the shared defaults, seeding goes through the repository, an operator's value is preserved, a re-run seeds nothing new, a failed seed does not fail tenant setup, and the shared-default lookup itself
- **4 e2e** — both ends readable over the API, the range is a real range, setup is re-runnable without disturbing the values

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |


kugel-masa:
## Code Review (PR #175)

Reviewed against the running stack. One gap, closed in e7fc26c.

### Findings

#### 🟡 The document that caused this confusion still described the old chain — fixed in e7fc26c

`docs/{en,ja}/master-data/model-specification.md` listed setting resolution as terminal → store → global default and stopped there. The fourth step — the consuming service's own configuration — is exactly what made this issue hard to see: the endpoint answers **404** rather than the value the service would actually use. Leaving the spec at three steps means the next reader repeats the same reasoning.

Now documents the fourth step, the seeding, and its consequence (below).

#### 🔵 Seeding turns an implicit default into an explicit record

After seeding, `RECEIPT_NO_START_VALUE` is a stored value for that tenant. If the shipped default ever changes, existing tenants keep what they were seeded with and only new tenants get the new value.

That is the right trade — a client-visible number should be explicit configuration, not something that shifts under a deployment — but it is a behaviour change worth stating rather than discovering, so it is in the spec now.

#### 🔵 Not changed

- **Concurrent tenant setup**: two simultaneous `POST /tenants` could both pass the existence check and race on the insert. The unique index on `(tenant_id, name)` decides it and the loser is logged as a warning, which is the correct outcome — tenant setup is not a hot path and the seeded value is identical either way.
- The seeding covers the two settings a terminal needs today. If more terminal-facing values appear, they go in `TERMINAL_FACING_SETTING_NAMES` rather than growing a general "seed everything with a default" mechanism, which would freeze every default at tenant-creation time.

### Verified during this pass

- **Ordering**: seeding runs after `create_collections`, so the settings collection and its unique index exist first.
- **Lookup semantics**: `get_settings_by_name_async` → `get_one_async` returns `None` when absent (it does not raise), so the insert-if-absent check does what it reads like — otherwise seeding would have been silently skipped by the exception branch. The re-run e2e test covers it end to end.
- **Defaults are single-sourced**: read from `AppSettings.model_fields`, not restated in this service, so they cannot drift from the values the services number with.
- **The 422 regression is genuinely fixed**: `GET /settings` returns 200 and lists the seeded rows alongside the tenant's own.
- `ruff check` clean on the changed files; master-data unit 394 pass.

### Testing

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |

### Note for whoever runs the suites next

master-data's `tests/integration/test_operations.py` asserts a store-item count that depends on `49-01` — data seeded by the **e2e** tier's `test_setup_data`. Reset the tenant database and it fails until the e2e setup runs again. Unrelated to this change, but it cost time to rule out; worth its own issue if the tiers are meant to be independent.

### Verdict

**Comment only** — the finding is closed in this branch. I authored the change, so this is not an independent approval.

kugel-masa:
### Follow-up: the seeded value now follows the deployment, not the constant

Answering "which default does it use when the setting is not registered" turned up a trap this PR would have introduced.

Two different defaults were in play:

| | source | environment-aware |
|---|---|---|
| runtime fallback (cart) | `getattr(settings, name)` → commons `AppSettings` | **yes** — pydantic-settings |
| the value this PR seeded | `AppSettings.model_fields[name].default` | **no** — the class constant |

Since a seeded record outranks a service's own configuration, a deployment that had raised the range by environment would have been silently reverted the moment its tenant was set up:

```
before seeding:  cart resolves RECEIPT_NO_START_VALUE=200000 from its environment
after seeding:   master holds 111111, which wins  ->  200000 quietly stops applying
```

Nothing in this repository sets those variables, so the shipped behaviour never changed — the defect was latent, and only visible by asking where the value comes from.

The resolver now instantiates `AppSettings(_env_file=".env")`, i.e. resolves the way a service resolves its own settings, and the model spec states the consequence: a deployment overriding by environment must set the same variable for master-data or configure the tenant setting directly. Two tests pin it — the resolver follows an environment override, and the seeding writes what the deployment configured.

kugel-masa:
## Code Review — second pass (after the environment-resolution change)

### Findings

#### 🟡 The seeding guard did not cover the resolution — fixed in the latest commit

`seed_terminal_facing_settings` wraps its work in a try so that a tenant which cannot be seeded is still created. Resolving the value moved into that function reading the environment, but sat **outside** the try:

```python
for name in TERMINAL_FACING_SETTING_NAMES:
    default_value = _shared_default(name)   # reads the environment, outside the guard
    ...
    try:
        ...
```

So a resolution failure would have taken tenant creation down with a 500, which is exactly what the guard exists to prevent.

How reachable is it? Less than it looks: a malformed variable stops the service at **import** time, because the commons settings instance is built there —

```
RECEIPT_NO_START_VALUE=not-a-number
  ValidationError: Input should be a valid integer  (kugel_common/config/settings.py:65, at import)
```

— so the process never gets as far as serving a tenant request. What remains is a `.env` that becomes unreadable or malformed while the service is running. Narrow, but the code states a guarantee and should keep it. Resolution is now inside the try, with a test.

#### 🔵 Not changed

- The e2e assertions pin the shipped defaults (`111111` / `999999`). A deployment that overrides them by environment would see those tests fail — correctly, since the point of the change is that seeding follows the deployment. Worth knowing before running the suite against a customised environment.
- `_shared_default` instantiates `AppSettings` per name, so tenant setup builds it twice. At tenant-creation frequency that is not worth caching.

### Verified during this pass

- **The override path works end to end**: `RECEIPT_NO_START_VALUE=200000` in the environment produces a seeded `200000`, and the untouched case still produces `111111`.
- **Type handling**: the value round-trips through the `AppSettings` int field, so a non-numeric override is rejected at startup rather than being seeded as a string.
- **No divergence between the two readers**: master-data's own settings object and the `AppSettings` instance read the same environment, so they cannot disagree about the value being seeded.
- master-data unit 397 pass, e2e 19 pass; full run: unit 8/8, integration 7/7, e2e 8/8.

### Verdict

**Comment only** — the finding is fixed in this branch. I authored the change, so this is not an independent approval.

kugel-masa:
### Tests added for the two gaps named in the review

**Integration coverage for the seeding** (`master-data/tests/integration/test_settings_seed.py`, 7 tests)

The unit tests mock the repository, so they pin the decisions — what is seeded, what is left alone — but not the document that actually lands. That distinction is not academic here: the first version of this change inserted documents directly, which looked correct in Mongo and then made `GET /settings` fail with a 422, because the response renders `entry_datetime` from `created_at` and only the repository stamps it. That bug was caught by an e2e run, one tier further out than it should have been.

The new tests exercise the real write and the real read-back: both documents exist with `created_at`, the list and value endpoints render them, a re-run neither duplicates nor overwrites a value an operator configured.

Checked that they earn their place — putting the raw insert back:

```
FAILED TestSeededDocuments::test_the_documents_carry_what_the_service_layer_needs
FAILED TestReadBackThroughTheApi::test_the_settings_list_renders          <- the 422
FAILED TestRerunningSetup::test_does_not_overwrite_a_configured_value
3 failed, 4 passed
```

**Cart's e2e asserted a constant** (`cart/tests/e2e/test_receipt_numbering.py`)

It compared the printed receipt number against `111111` written in the test file. That passes whether or not cart and the client resolve the *same* range — which is the thing this issue is about. The suite now reads the range from master-data exactly as a client would and asserts cart numbers with the same one, so it checks the agreement rather than a coincidence, and follows a tenant that overrides the range instead of failing spuriously.

unit 8/8, integration 7/7, e2e 8/8 suites pass.

### Still not covered

A test proving cart uses the *seeded record* rather than its own identically-valued default would need a non-default range to take effect mid-suite, which #176 currently prevents (the invalidation endpoint reports success without clearing these entries, and the TTL is 600s). Worth revisiting once #176 is fixed.

kugel-masa:
## External review (Gemini CLI 0.56.0, gemini-3-flash)

Ran the diff past a second model to get a view that is not mine. Five findings, all rated minor by it; I verified each against the code rather than taking them as read.

### Acted on

**The e2e range fixture masked exactly the failure it was meant to catch** — fixed in 26d8a63.

> *"The `receipt_range` fixture catches all exceptions and silently falls back to `SHIPPED_START`/`SHIPPED_END`. If the seeding fails or the API returns a 404, the e2e tests will still pass by using the internal fallback values, failing to verify the actual integration issue (#174) this PR aims to solve."*

Correct, and worse than stated. Checking it turned up that the fallback was firing **on every run**: the request carried `X-API-KEY` but no `terminal_id`, which is what that credential is checked against, so the endpoint answered **401** and the fixture quietly used its constants. The cross-service assertion I added in the previous commit was never crossing services. The request now sends `terminal_id`, and a non-200 fails the suite with an explanation instead of falling back.

**`AppSettings` rebuilt per setting** — now resolved once per seeding run, with the failure still guarded so a tenant that cannot be seeded is still created.

**Seeding does not say where its value came from** — the log line now distinguishes "environment" from "shipped default". A deployment that overrides the range for one service and not master-data gets the default seeded, and that record then outranks the service's own setting; the log makes the mismatch visible rather than puzzling.

### Assessed and not changed

**Check-then-act race between concurrent tenant setups.** Real, and already noted in my earlier pass. The unique index on `(tenant_id, name)` decides it and the loser is logged. The reviewer's added point — that the warning misleads for an expected concurrency event — is fair, but tenant setup is not concurrent in practice here and the seeded value is identical either way; catching `DuplicateKeyException` specifically would be tidier without changing any outcome.

**Broad `except Exception` around the seeding hides infrastructure failures.** Deliberate: a tenant that cannot be seeded still works (services fall back to their own configuration), and failing tenant creation over a default would be worse. A dead database fails earlier anyway, in collection creation.

### Note

The free tier rate-limited the first two attempts (5 requests/minute on `gemini-3-flash`); it succeeded on the third. Worth knowing before leaning on this in a loop.

unit 8/8, integration 7/7, e2e 8/8 suites pass.

kugel-masa:
## Verified through the operator path, not just the seeded one

The earlier verification set the range by environment variable, which exercises the seeding but not the path an operator would actually use. Repeated it through the settings API against the running stack.

**Set a custom range** (deliberately tiny, so a wrap is reachable):

```
PUT /api/v1/tenants/T6216/settings/RECEIPT_NO_START_VALUE  {"defaultValue": "500000"}  -> 200
PUT /api/v1/tenants/T6216/settings/RECEIPT_NO_END_VALUE    {"defaultValue": "500004"}  -> 200
GET .../RECEIPT_NO_START_VALUE/value  -> 200 {"value": "500000"}
```

**cart numbers with it** (terminal-JWT authenticated, carried finalize):

```
receiptCounter=1  ->  receiptNo=500000     first of the range
receiptCounter=5  ->  receiptNo=500004     last of the range
receiptCounter=6  ->  receiptNo=500000     wrapped
receiptCounter=7  ->  receiptNo=500001
```

**Persisted and propagated**:

```
cart log_tran:  seq=711 counter=1 receipt_no=500000  … 4 rows
journal: 4 rows   report: 4 rows
```

So a value configured through the documented API reaches the finalize path and the range cycles at the configured end — which is what this issue was about, from the side an operator sees.

### One caveat, tracked separately

cart only picked the new range up after a **restart**. Updating the setting and calling `DELETE /api/v1/cache/master-data?namespace=settings_master` (200) left it numbering from the old range. That is #176, and this is its second reproduction — the first was through the seeding path, so it is not specific to a freshly seeded record: an ordinary settings update is equally invisible until the process restarts.

For an operator that reads as: the settings API confirms the new range, and every receipt keeps printing the old one.

### Environment restored

Range set back to 111111 / 999999, probe transactions removed from cart / journal / report, stack restarted, e2e 8/8 green.

kugel-masa:
### Correction to the caveat above

The note I added about cart only picking up a changed range after a restart was wrong, and #176 has been corrected accordingly.

Cache entries and their generation counter are scoped by **(tenant, store, namespace)**, and a terminal's settings repository is store-scoped:

```
cart||mdcache:T6216:5678:settings_master:gen0:one:RECEIPT_NO_START_VALUE
```

I invalidated **without `store_code`**, which bumps the tenant-scoped generation while those entries read the store-scoped one — so the call did exactly what I asked and nothing else. With the store code it takes effect immediately, no restart:

```
master set to 600000..600003
DELETE /api/v1/cache/master-data?namespace=settings_master&store_code=5678  -> 200
  receiptCounter=1 -> receiptNo=600000
  receiptCounter=4 -> receiptNo=600003     (range end)
  receiptCounter=5 -> receiptNo=600000     (wrapped)
```

The "restart fixed it" reading was the 600s TTL expiring across my experiments.

So the operator path works end to end without qualification: set the range through the settings API, invalidate the namespace **for the store**, and the next cart numbers with it. What remains in #176 is only that an invalidation of the wrong scope reports success like any other.

Environment restored to 111111 / 999999; probe transactions removed.

PR #177, opened 2026-08-20
fix(cart,commons): recover the loser of a concurrent finalize (#172)

### LINKED ISSUE #172: cart: two concurrent identical finalizes return 500 to the loser (abort after commit masks the real error)
Found while reviewing #171. **Pre-existing — reproduces on `bill` without any of that PR's changes.**

## Problem

Two identical finalizes issued concurrently (the shape a POS client with a short timeout produces) leave the data correct but hand one caller a 500:

```
A -> 200 {"success":true, ...}
B -> 500 {"code":900999, "data":"Cannot call abortTransaction after calling commitTransaction"}
tranlogs for cart: 1
```

Reproduced against the running stack on both `POST /carts/{cart_id}/bill` and `POST /carts/{cart_id}/cancel`, with `asyncio.gather` of two identical requests.

## Why

`tran_service.create_tranlog_async` runs the finalize in a manual session (`tran_service.py:311`) and has a dedicated recovery for the race:

```python
except DuplicateKeyException as e:
    # bug_001: a concurrent identical finalize won the race ...
    await self.tranlog_repository.abort_transaction()
    ... re-read the winning tranlog and return it idempotently
```

That branch catches a duplicate raised by the **insert**. In the concurrent case the loser's insert is buffered and the unique-index violation surfaces at **commit** instead, so `commit_transaction()` raises and control falls into the generic handler:

```python
except Exception as e:
    await self.tranlog_repository.abort_transaction()      # session已 ended by the failed commit
    message = f"Error creating tranlog: {e}"
    raise InternalErrorException(message, logger) from e
```

Aborting a session whose commit already ran raises `Cannot call abortTransaction after calling commitTransaction`, which **replaces the original error** — so the log does not even say what really failed, and the client sees 900999 instead of the idempotent result the winner already wrote.

## Impact

The `cart_id` dedupe (#152) makes a **sequential** lost-ACK retry converge — that is covered and passes. A **concurrent** retry does not: the operator sees a failure for a sale that was in fact recorded exactly once, and a client that reacts by retrying again just repeats the race.

## Suggested fix

1. Make the abort conditional (only abort a session that is still in a transaction), so a failed commit cannot mask its own cause.
2. Recognise a duplicate-key failure raised at commit time and route it to the existing idempotent recovery, so the loser returns the winner's tranlog like the insert-time path already does.
3. A regression test for the concurrent case, alongside the existing sequential ones.

## Reproduction

```python
req = {"signedSnapshot": snap, "payload": {"seq": N, "receiptCounter": M, "transactionDatetime": "..."}}
a, b = await asyncio.gather(client.post(url, json=req), client.post(url, json=req))
# one 200, one 500; exactly one tranlog persisted
```

## References

- #152 / #156 — the cart_id dedupe this defeats for concurrent callers
- `services/cart/app/services/tran_service.py:311-365` — the session block, the bug_001 recovery, and the generic handler
- #171 — where this surfaced (a cancellation now takes the same finalize path, so it inherits the behaviour)

kugel-masa:
Closes #172.

## Problem

Two identical finalizes in flight — the shape a POS with a short timeout produces — left the data correct and handed one caller a 500 whose message described a cleanup error:

```
A -> 200
B -> 500  "Cannot call abortTransaction after calling commitTransaction"
tranlogs for cart: 1
```

Reproduced on both `bill` and `cancel` with `asyncio.gather` of two identical requests.

## Two faults, stacked

**`abort_transaction` raised.** Callers abort from an exception handler, so an abort that raises replaces the failure being handled and loses the reason — the log did not even say what really went wrong. A session whose commit has run is past the point where it can be aborted and the driver says so. It now aborts only while the session is still in a transaction, logs anything that goes wrong rather than propagating it, and releases the session either way.

**The recovery did not cover the shape the race actually takes.** bug_001 handles a `DuplicateKeyException` from the insert. In the concurrent case the duplicate arrives *wrapped* — `CannotCreateException`, "Failed to save document to database" — so it reached the generic handler, which had no recovery at all. Both paths now go through one:

```python
existing = await self.__recover_concurrent_finalize(tranlog)
if existing is not None:
    return self.__apply_tranlog_to_cart(cart, existing)
```

abort → drop the poisoned session → ask whether a concurrent request already wrote this finalize. If it did, return it (the winner already published, so we do not publish again). A cart_id belonging to a **different** finalize still raises `FinalizeConflictException` (bug_008); a failure with no persisted counterpart still surfaces as an error.

The lookup moved into `TranlogRepository.get_existing_finalize_async`, which is the same check the insert pre-check runs — idempotency is now decided in one place instead of two.

## A third defect the fix uncovered

With the 500 out of the way, the loser's response failed validation with `receipt_no=None`. The recovery returned the winner's tranlog but skipped copying its numbers onto the cart the response is built from — the early `return existing` had always skipped it, invisible while nobody reached that path. Factored into `__apply_tranlog_to_cart`, used by the normal and the recovery path alike.

## Result

```
A -> 200   transactionNo=9501  receiptNo=111151
B -> 200   transactionNo=9501  receiptNo=111151
tranlogs for cart: 1
```

with the recovery visible in the log:

```
Concurrent finalize race for cart_id=…; returning the winning tranlog idempotently
Finalize failed but the same finalize is already persisted (cart_id=…, CannotCreateException); returning it
```

That log line is deliberately terse — the exception carries the whole tranlog, and one race must not put a full cart document in the log (#155).

## Testing

- **6 commons unit** — aborts a live transaction, skips the abort once the transaction has ended, never raises (refused abort, failed close), releases the session anyway, and still rejects an abort with no transaction at all
- **6 cart unit** — recovery returns the winner, returns nothing when the failure was something else, aborts before looking up, still conflicts for a different operation, and is reached from both failure shapes
- **2 e2e** — two identical bills and two identical cancels raced with `asyncio.gather`: both callers get 200 with the same numbers, and exactly one tranlog exists

The e2e module ensures tenant setup first: the unique index on `(tenant_id, store_code, cart_id)` is what stops the second insert, and without it the test would be measuring an unconfigured database.

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |

kugel_common 0.1.71.

### DIFF
--- a/services/account/Pipfile
+++ b/services/account/Pipfile
@@ -13,7 +13,7 @@ motor = "*"
 pytz = "*"
 pyjwt = "*"
 passlib = {extras = ["bcrypt"], version = "*"}
-kugel_common = {file = "commons/dist/kugel_common-0.1.70-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.71-py3-none-any.whl"}
 
 bcrypt = "==3.2.0"
 python-multipart = "*"
--- a/services/account/Pipfile.lock
+++ b/services/account/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "970532409b4de81a77b486cf57de3db140a3ac934bf1e803b1ee1cf24013de27"
+            "sha256": "a3319c2aee41f560546091e5c789b0d32c4715bc3d120d183917dfb97a039d1b"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -667,9 +667,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.70-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.71-py3-none-any.whl",
             "hashes": [
-                "sha256:4d00f28e41767914b0868c20af66d529ae563716590b54d3579950d070ff7e0a"
+                "sha256:147242bc32e62ed21ad52aa140cb7fd27014bce87c7d5a86ddd791682256581d"
             ]
         },
         "motor": {
--- a/services/cart/Pipfile
+++ b/services/cart/Pipfile
@@ -12,7 +12,7 @@ pydantic-settings = "*"
 motor = "*"
 pytz = "*"
 pyjwt = "*"
-kugel_common = {file = "commons/dist/kugel_common-0.1.70-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.71-py3-none-any.whl"}
 httpx = "*"
 aiohttp = "*"
 wcwidth = "*"
--- a/services/cart/Pipfile.lock
+++ b/services/cart/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "85eabb46c8a819c1cd200de0fc18ab609f246d654d9942adcdeaf58ce4d1bb75"
+            "sha256": "a3989da3859d8b6aa748ef160fee7319eae028400ac2184ad9c5877a72b4aa18"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -659,9 +659,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.70-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.71-py3-none-any.whl",
             "hashes": [
-                "sha256:4d00f28e41767914b0868c20af66d529ae563716590b54d3579950d070ff7e0a"
+                "sha256:147242bc32e62ed21ad52aa140cb7fd27014bce87c7d5a86ddd791682256581d"
             ]
         },
         "motor": {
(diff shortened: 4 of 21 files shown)


kugel-masa:
## External review (Gemini CLI, gemini-3.6-flash)

One finding, rated major, and it is correct.

### Acted on — fixed in af59d2f

> *"The unit tests `test_duplicate_at_insert` and `test_failure_at_commit` … call the private recovery helper `__recover_concurrent_finalize` directly rather than the SUT method `create_tranlog_async`. This means they bypass the exception-handling blocks entirely, so they cannot verify that `create_tranlog_async` catches the exceptions and routes them to recovery correctly."*

Exactly right, and worth spelling out: those two tests were named for the property this PR exists to establish — that **both** failure shapes reach the recovery — while testing the helper that the routing calls. With the routing deleted they would still have passed. They were also, once the arranged-but-unused exception is ignored, duplicates of `test_returns_the_winners_tranlog`.

They now drive `create_tranlog_async` itself, with the failure arranged at the insert (`DuplicateKeyException`) and in the wrapped form the race actually produces (`CannotCreateException`), asserting what a caller observes: the winner's tranlog returned, its numbers copied onto the cart the response is built from, and **no second publish**. A third case pins the other direction — a failure with nothing persisted still raises.

Verified they now earn their place. Removing the recovery from the generic handler:

```
FAILED tests/unit/.../TestReachedFromBothFailureShapes::test_failure_at_commit
1 failed, 6 passed
```

The previous version stayed green through the same mutation.

### The rest of the review

No critical or minor findings. The reviewer worked through the questions I put to it and reached the same conclusions I had, with one addition I had not checked and should have: the loser skipping the publish is safe **because the delivery-status record is a transactional outbox** — the winner's record is committed with its tranlog, and `republish_undelivered_tranlog_async` in `cron/republish_undelivery_message.py` republishes anything left pending. So even if the winner's direct publish fails, the event still goes out; the loser has nothing to compensate for. That is the answer to "could downstream miss the transaction entirely", and it comes from the design rather than from this change.

It also confirmed the loser's own delivery-status write is rolled back with the aborted transaction, so no orphan `event_id` is left behind.

unit 8/8 suites pass; integration and e2e unchanged by a test-only commit (both green on the previous one).

PR #178, opened 2026-08-20
feat(cart): detect a finalize that falls into the other numbering series (#168)

### LINKED ISSUE #168: cart: DUAL mode has two independent receipt-number sources — a degraded snapshot silently reissues numbers

kugel-masa:
Closes #168. The last of the three backend items the stateless migration needs.

## Problem

DUAL mode keeps **two independent receipt-number sources**, and the branch is per *transaction*:

| request | numbers come from |
|---|---|
| carries a finalize context | the terminal's running `receipt_counter` |
| carries none | cart's own `terminal_counter` collection |

A phase 2 terminal whose snapshot signing has degraded — `SNAPSHOT_HMAC_KEYS` unset, malformed, or mid-rotation — is issued no snapshot, so it cannot carry anything and its next sale is numbered from the other series. That series knows nothing about how far the terminal has advanced, so it can print a receipt number the terminal has already issued.

Observed on the running stack while verifying this change:

```
carried bill        -> receiptNo 111170   (counter 60)
snapshot-less bill  -> receiptNo 111160   (the other series, which will reach 111170 in time)
```

## Detected, not blocked

Refusing the finalize would stop a store selling over a key misconfiguration, and the posture for numbering integrity here is audit detection rather than enforcement (spec 156 Q58). `CART_REQUEST_SNAPSHOT_MODE=REQUIRED` is what removes the window — a snapshot-less mutating request is rejected, so there is no second series to fall into.

Until then, a finalize that will be server-numbered **while the terminal has a series of its own** logs at ERROR:

```
Finalize numbered from the server-side series while the terminal has its own
(issue #168): cart_id=… reason=no_carried_context terminal_receipt_counter=60
stateless=False. Receipt numbers from the two series can collide; fix the
snapshot signing key, or move to CART_REQUEST_SNAPSHOT_MODE=REQUIRED.
```

and writes a `numbering_fallback` record to the restore audit trail with `reject_reason` naming which condition fired (`signing_degraded` / `no_carried_context`). A phase 1 terminal — no second series, nothing to collide with — is not flagged.

## Two gaps left over from #166, closed here

Knowing whether a terminal numbers its own receipts turned out not to be possible:

- the terminal token's `receipt_counter` claim was never mapped back into `TerminalInfoDocument` (the claim was added in #166, the reverse mapping was not)
- the terminal service's detail response did not expose the field at all, so a service reached with an **API key** — which has no claims to read — could not see it either

Both are fixed, which is also what made the first version of this detection silently never fire. Generated API docs regenerated for the response change.

## Testing

- **7 unit** — fires for a terminal with its own series and for degraded signing; stays quiet for a phase 1 terminal and for a counter of 0 (seeded at open, never advanced); survives a missing audit repository; and says the issue number and the remedy in the log
- **verified live** — the ERROR line and the `numbering_fallback` audit record both appear for a snapshot-less finalize from a terminal carrying counter 60

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |

kugel_common 0.1.72.

### DIFF
--- a/docs/en/cart/client-carried-cart.md
+++ b/docs/en/cart/client-carried-cart.md
@@ -183,6 +183,40 @@ Pre-#166 clients keep working: they carry no counter, send their number under th
 old `receipt_no` name at open (numerically the same value — they counted 1, 2, 3
 with no wrap), and their receipt numbers are recorded as sent.
 
+## Two numbering series while DUAL mode is on (#168)
+
+The finalize path branches per **transaction**, not per terminal:
+
+| request | numbers come from |
+|---|---|
+| carries a finalize context | the terminal's running `receipt_counter` |
+| carries none | cart's own `terminal_counter` collection |
+
+A phase 2 terminal whose snapshot signing has degraded — `SNAPSHOT_HMAC_KEYS`
+unset, malformed, or mid-rotation — is issued no snapshot, so it cannot carry
+anything and its next sale is numbered from the *other* series. That series knows
+nothing about how far the terminal has advanced, so it can print a receipt number
+the terminal has already issued.
+
+This is **detected, not blocked**: refusing the finalize would stop a store
+selling over a key misconfiguration, and the posture for numbering integrity here
+is audit detection rather than enforcement (spec 156 Q58). A finalize that falls
+into the server-side series while the terminal has one of its own logs at ERROR
+
+```
+Finalize numbered from the server-side series while the terminal has its own
+(issue #168): cart_id=… reason=signing_degraded terminal_receipt_counter=…
+```
+
+and writes a `numbering_fallback` record to the restore audit trail
+(`log_cart_restore`), with `reject_reason` naming which condition fired
+(`signing_degraded` / `no_carried_context`).
+
+`CART_REQUEST_SNAPSHOT_MODE=REQUIRED` removes the window entirely — a
+snapshot-less mutating request is rejected, so there is no second series to fall
+into. Until then, a degraded signing key is an incident to fix, and now one that
+is visible per transaction rather than only in the startup log.
+
 ## Configuration
 
 | Variable | Default | Purpose |
--- a/docs/en/cart/client-carried-cart.md
+++ b/docs/en/cart/client-carried-cart.md
@@ -209,8 +209,14 @@ Finalize numbered from the server-side series while the terminal has its own
 ```
 
 and writes a `numbering_fallback` record to the restore audit trail
-(`log_cart_restore`), with `reject_reason` naming which condition fired
-(`signing_degraded` / `no_carried_context`).
+(`log_cart_restore`), with `reject_reason` naming which condition fired (`signing_degraded`,
+`snapshot_without_finalize_context`, or `no_carried_context`).
+
+One case is deliberately out of reach: a phase 2 terminal that has not numbered
+anything yet, whose signing is healthy and which carries no snapshot at all,
+looks exactly like a phase 1 terminal. Open seeds every terminal's counter with
+zero, so treating zero as "has its own series" would report every legacy
+finalize as an incident instead.
 
 `CART_REQUEST_SNAPSHOT_MODE=REQUIRED` removes the window entirely — a
 snapshot-less mutating request is rejected, so there is no second series to fall
--- a/docs/en/terminal/api-specification.md
+++ b/docs/en/terminal/api-specification.md
@@ -785,6 +785,7 @@ Optional filtering by store code is supported.
 | `businessDate` | string | No | - |
 | `openCounter` | integer | Yes | - |
 | `businessCounter` | integer | Yes | - |
+| `receiptCounter` | integer | No | - |
 | `initialAmount` | number | No | - |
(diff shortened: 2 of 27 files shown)


kugel-masa:
## External review (Gemini CLI, gemini-3.6-flash)

Five findings — one critical, one major, three minor. Checked each against the code and the live data; the critical does not hold, the major does.

### 🔴 Critical — does not hold

> *"Legacy (phase 1) terminals have `terminal.receipt_counter` seeded unconditionally at open … any active phase 1 terminal with historical transactions has `receipt_counter > 0` … every single legacy finalize logs an ERROR and writes an audit record."*

The premise is wrong: **cart's server-side numbering never touches the terminal document**. `terminal.receipt_counter` moves only when a client *presents* a counter at open, which is precisely the phase 2 signal. A phase 1 terminal stays at zero however many transactions it books.

Checked against the live database rather than by reading:

```
terminals with receipt_counter > 0 : 0
terminals at 0 or unset            : 383
```

— including the many that have booked legacy transactions in test runs. So the noise the finding predicts does not occur.

### 🟡 Major — real, fixed in 4261049

> *"A newly opened phase 2 terminal starts with `receipt_counter = 0`. If its first transaction falls back while snapshot signing is functional, `bool(0)` is False. The fallback is silently ignored."*

Correct. But its suggested fix — test `is not None` — would produce exactly the false positives its own critical warns about, because open seeds **every** terminal with zero, phase 1 included.

Closed from the other side instead: a request that **carried a snapshot but no finalize context** is a phase 2 client whatever its counter says, so that is now a signal in its own right (`reject_reason=snapshot_without_finalize_context`).

One case is still out of reach, and now says so in the code and the docs rather than being left to be discovered: a phase 2 terminal at counter zero, with healthy signing, carrying no snapshot at all, is indistinguishable from a phase 1 terminal.

### 🔵 Minor

**"No test verifies that `bill_async` actually invokes the audit"** — correct and now fixed. Only the helper was covered, so deleting the call from `bill_async` would have left every test green. Verified the new test by deleting it: it fails.

**"`test_a_zero_counter_is_not_a_series` codifies the defect as correct behaviour"** — half right. It pins a deliberate trade-off, not a defect, but the test said nothing about *why*. Renamed and rewritten to state it, with the snapshot-signal test next to it.

**"The audit write is awaited on the finalize path and could slow a sale"** — kept as is. It only runs on the incident path: a healthy phase 1 deployment never reaches it, and a phase 2 one reaches it only while signing is broken. Making it fire-and-forget would trade a bounded cost on an already-broken path for losing the record that says the incident happened.

### Not changed

The reviewer confirmed the numbering behaviour itself is untouched (observation only), that the audit failure path cannot break a sale, and that exposing `receipt_counter` raises no security, tenancy or compatibility concern.

unit 8/8, e2e 8/8 suites pass.

PR #179, opened 2026-08-20
feat(cart): monotonic snapshot revision so a rollback is visible (#165)

### LINKED ISSUE #165: cart: the signed snapshot has no freshness — an older validly-signed envelope is accepted (add a monotonic revision for rollback detection)

kugel-masa:
Closes #165.

## Problem

The signature proves an envelope was issued unmodified. It does not prove it is the **current** one — and on the stateless path the server deliberately does not consult the cache, so it has no other basis for telling a current envelope from an earlier one for the same cart.

Refusing a stale envelope synchronously would mean knowing the high-water mark per cart, i.e. a per-request write — exactly what phase 2 was built to remove. So this makes rollback **findable**, not preventable.

(The finalize side is already covered: #152's `cart_id` dedupe means resurrecting a pre-finalize envelope cannot produce a second transaction. The gap is rollback *before* finalize, which leaves no trace at all today.)

## What it does

The cart document carries a monotonic `revision`, covered by the signature and advanced once per issued snapshot:

```
create → revision 1 → add item → revision 2 → add item → revision 3
```

The revision a request **presented** is recorded on its request log entry, as its own field rather than in the body — which the logging middleware strips (#155):

```json
"snapshot_info": {"cart_id": "…", "revision": 3, "schema_version": 2, "kid": "v1"}
```

Normal operation is strictly increasing; a lost-ACK retry repeats a value; a replayed older envelope is the one that goes **down**. An index on `(tenant_id, snapshot_info.cart_id, request_info.accept_time)` makes that a query rather than a hunt.

Envelopes are issued at `schema_version` 2. Version 1 stays accepted — a client that has not migrated presents one, and refusing it would break the failover the snapshot exists for; it simply carries no revision to record.

## Two mechanics worth knowing

**The peel runs outside the logger.** The middleware that peels the envelope is registered outside `log_requests`, so by the time the request is logged the envelope is gone. The peel leaves the scalars on the request scope under a key commons reads (`request_log_snapshot`) — which is also why `revision` had to be a dedicated `RequestLog` field rather than something read back out of the body.

**An envelope arrives in the wire form.** A terminal presents what it received — camelCase, per the response alias generator — not the snake_case signing canonical form. The extractor accepts both. The first version read only snake_case and silently recorded nothing; caught by running it against the live stack rather than by reading it.

## Testing

- **13 unit (cart)** — issuing advances the revision and continues from what the client carried, the signature covers it, both schema versions verify, marks extraction over the wire form and over malformed input
- **6 unit (commons)** — the scope pickup, a request carrying nothing, a v1 envelope with no revision, and bad input costing only the marks rather than the whole log entry
- **4 integration** — against the real request log with the buffer flushed on demand, including a replay showing up as a lower revision and ordinary use showing none
- **2 e2e** — over the wire: revisions advance and are strictly increasing, a replayed envelope is accepted and the response continues from *its* revision

| tier | result |
|---|---|
| unit | 8/8 suites pass |
| e2e | 8/8 suites pass |

kugel_common 0.1.73.


kugel-masa:
## Review follow-up (external review by Gemini + verification)

Three findings acted on, all in `snapshot_service.py` — no commons redistribution.

### The marks came off an *unverified* envelope, unbounded

Extraction runs before verification, and `log_requests` writes its record in a
`finally`, so a rejected request is logged just the same. Nothing capped what
`snapshot_info` could carry — which reopened, through a new field, the log
amplification #155 closed on the body. Measured against the running stack with a
5.1 MB request carrying a 5,000,000-char `cart_id` and a 100,000-char `kid`:

```
before   snapshot_info alone serialized to 5,100,062 bytes
after    HTTP 400, and the whole log_request document is 1,262 bytes
         snapshot_info: {'cart_id': None, 'revision': 3, 'schema_version': 2, 'kid': None}
```

Over-length strings are **dropped, not truncated** — a `cart_id` is a uuid4, and
a truncated one would read as real while matching no cart. The cap is 128 chars,
the same one the body strip already applies to metadata it carries over
(`_STRIP_META_MAX_STR`).

### `bool` passed the int screen

`isinstance(x, int)` is `True` for `True`. A `revision: true` was recorded as
revision 1 — a mark the envelope never carried. Now `type(x) is int`.

### A non-string `cart_id` cost the whole record

`RequestLog.SnapshotInfo(cart_id={...})` raises `ValidationError`, and
`_make_snapshot_info` drops the entire record when construction raises — so the
malformed envelope, the case a rejection most needs traceable, was the one that
recorded nothing. The `revision` and `kid` alongside it were lost too. Screened
at extraction instead, so only the bad field is dropped.

`extract_audit_meta` reads the same unverified values into the restore audit
trail and now shares all three screens.

### Not changed: the cache path does not persist the bump

Raised as critical in review; downgraded after checking. The revision advances
when a snapshot is *issued*, by which point the cache-authoritative path has
already written the cart — so the bump reaches the client but not the cache. A
cart that switched back to the cache path mid-life would present a lower
revision and read as a rollback.

That switch is not a case to tune the detection for: it rebuilds the cart from a
cache the stateless path deliberately stopped writing, so the **cart itself** has
rolled back, not just its revision — the pre-existing consequence of #156, and
far louder than a log anomaly. All 12 cart routes are body-carrying mutations, so
a client that carries carries throughout. Documented rather than coded around.

### Verification

- New guards mutation-checked: reverting each cap/screen fails 5 of the new tests
- The revision increment itself mutation-checked: 3 existing tests fail without it
- cart unit 637 passed · integration 52 passed · e2e 72 passed · commons unit 529 passed
- Live probe above run against the rebuilt stack under terminal-JWT auth

kugel-masa:
## Test coverage follow-up

Two gaps found reviewing the coverage of the previous commit, both closed in `395ebd0`.

**The new bounds were unit-tested only.** The unit tier proves what the function returns; the bound exists for what MongoDB ends up holding. Added an integration test that sends a rejected request carrying a 1 MB `cart_id` and a 100 KB `kid`, then measures the stored document:

```
assert len(bson.BSON.encode(row)) < 64 * 1024
assert marks["cart_id"] is None and marks["kid"] is None
assert marks["revision"] == 3 and marks["schema_version"] == 2   # still traceable
```

**Nothing asserted the index exists.** The docs say the rollback query has one; if the spec were dropped, the query would silently become a collection scan and no test would notice. Added an assertion that it exists and is partial.

Both mutation-checked — removing the cap fails the first, removing the index spec fails the second.

**Also documented an operational step that was missing.** The index is created by tenant setup and by nothing at startup, so on an environment that already has tenants, `POST /api/v1/tenants/{tenant_id}` must be re-run once per tenant before the query is indexed. Index creation is idempotent, and this is the same step #156 needed — but without saying so, "there is an index for it" is not true of an upgraded deployment.

### Left uncovered, deliberately

- **Concurrency.** Two requests presenting the same revision both receive the same next value. Inherent to a stateless design with a single-threaded terminal, and a duplicate revision reads as non-increasing, which is arguably the correct signal for a forked envelope.
- **Path mixing.** The cache-path revision anomaly is documented rather than tested; the scenario rolls the cart itself back, so a test would be asserting on already-broken behaviour.

### Dependency worth stating

This trail is only as good as the request log under it. #180 (idle-flush loss) was hit for real while verifying this PR — a single carried request did not reach MongoDB until 110 filler requests pushed the buffer past its size trigger. A quiet lane loses the rollback evidence along with everything else.

### Totals

cart: unit 637 · integration 54 · e2e 72 · commons: unit 529 — all green, tiers run separately per the project's runners.

(Unrelated: running `-m "unit or integration"` in one process fails 11 log-assertion tests. Identical on `main`, so pre-existing cross-tier logging pollution, not from this PR.)

Issue #182, opened 2026-08-21
commons: the request-log unique index keys on fields that do not exist, so it constrains a tenant-wide timestamp instead of a terminal


kugel-masa:
Every service creates its `log_request` collection with this index:

```python
# services/*/app/database/database_setup.py
{"keys": {"tenant_id": 1, "store_code": 1, "terminal_no": 1, "request_info.accept_time": 1},
 "unique": True}
```

A request log document has no `store_code` or `terminal_no` at the top level — they live under `terminal_info`:

```
_id, shard_key, created_at, updated_at, etag, tenant_id, client_info,
request_info, response_info, staff_info, user_info, terminal_info, snapshot_info
```

Missing fields index as null, so for every document the key is `(tenant_id, null, null, accept_time)`. The effective constraint is **"one request per tenant per timestamp"** — not per terminal, which is what the index was written for.

## Demonstration

Two requests from **different terminals** in the same tenant, same `accept_time`:

```js
c.insertMany([
  {tenant_id:"T6216", terminal_info:{store_code:"5678", terminal_no:9}, request_info:{accept_time:t}},
  {tenant_id:"T6216", terminal_info:{store_code:"9999", terminal_no:1}, request_info:{accept_time:t}},
], {ordered:false})

→ write errors: 1
→ code: 11000 (duplicate key)
→ 1 of 2 inserted
```

The second terminal's entry is rejected.

## Consequences

- **The index does not do its job.** It was given per-terminal separation and has never had it, under any credential.
- **A collision drops the tenant-DB copy only.** The buffer writes each log to `{prefix}_commons` and `{prefix}_{tenant}`, and only the per-tenant collection carries this index (`create_request_log_collection` is called per tenant). So a collision leaves the two copies silently diverged rather than losing the record outright.
- **The error line overstates it.** `RequestLogBuffer._flush_unlocked` catches `BulkWriteError` and logs `Failed to flush {len(docs)} request logs`, when `ordered=False` means only the rejected documents failed.

## How likely

`accept_time` is `datetime.isoformat()` — microsecond precision — so a collision needs two requests in the same microsecond. That is rare, but the collision domain is now a whole tenant rather than one terminal, so it grows with fleet size rather than staying at zero. No collisions appear in current local data (checked by grouping on `accept_time`); this is filed for the constraint being wrong, not for observed loss.

## Suggested fix

Point the index at the fields that exist:

```python
[...]

Issue #181, opened 2026-08-20
commons: a terminal-JWT request logs no store or terminal, so the request log loses attribution as clients migrate


kugel-masa:
Found while verifying #165 against the running stack.

## Observed

Two requests to the same cart endpoint, same terminal, differing only in how they authenticated. Both land in the tenant database — the terminal token carries `tenant_id`, so `verify_token` resolves it through the user path — but they record different things:

```
API key      tenant_id=T6216
             terminal_info={"tenant_id":"T6216","store_code":"5678","terminal_no":9,
                            "business_date":"20260821","open_counter":1}
             user_info={"username":""}

terminal JWT tenant_id=T6216
             terminal_info={"tenant_id":"","store_code":"","terminal_no":0,
                            "business_date":"","open_counter":0}
             user_info={"username":"terminal:T6216-5678-9"}
```

## Why

`_get_terminal_info` (`log_requests.py`) resolves the terminal from `X-API-Key` plus a `terminal_id` query/path parameter. A JWT request has neither, so it returns None and `_make_terminal_info` writes the empty document — even though the token's claims carry `store_code`, `terminal_no`, `business_date` and `open_counter`, and `_terminal_info_from_claims` already knows how to build a `TerminalInfoDocument` from them.

## Impact

The JWT flow is the documented direction (`docs/ja/terminal/frontend-jwt-migration-guide.md`: "cart は `terminal_id` クエリも `X-API-KEY` も不要になります"), so as clients migrate the request log stops carrying store/terminal attribution altogether:

- operational queries by store, terminal or business date stop working; the information survives only as text inside `user_info.username` (`terminal:{tenant}-{store}-{terminal_no}`)
- `staff_info` is emptied too. It is resolved from the same terminal document, so a JWT request records `{"id": "", "name": ""}` where an API-key request records the signed-in staff. Measured side by side on the same request:

  ```
  JWT      staff_info={"id": "", "name": ""}
  API key  staff_info={"id": "S001", "name": "Staff1"}
  ```

- the rollback trail added in #165 is unaffected in itself — `snapshot_info` and `tenant_id` are identical under both credentials, so the detection query still works. What stops working is the step after it: a rollback can be traced to a cart, and to a terminal only by parsing the username string, but never to a business date or open session.

~~`log_request` has a unique index on `(tenant_id, store_code, terminal_no, request_info.accept_time)`. For JWT requests that degrades to `(tenant_id, "", 0, accept_time)`~~

[...]

PR #183, opened 2026-08-21
fix(commons): stop the request-log buffer from losing entries (#180)

### LINKED ISSUE #180: commons: the request-log buffer silently DROPS entries whenever the idle timer flushes
Found while verifying #165 against the running stack: request logs for a handful of requests never appeared in MongoDB, while bursts of a hundred did.

## The defect

`RequestLogBuffer` flushes on two triggers: the buffer reaching `max_size` (100) and an idle timer (5s). **The idle one loses the batch.**

`_flush_unlocked` takes the entries out of the buffer and then cancels the timer task:

```python
to_write = self._buffer
self._buffer = []                      # entries are now only in to_write
...
if self._timer_task and not self._timer_task.done():
    self._timer_task.cancel()          # <- when called FROM the timer, this cancels the running task
```

When the flush was triggered by `add()` that is correct — a different task is cancelled. When it was triggered by `_timer()` itself, `self._timer_task` **is** the running task, so it cancels itself. The cancellation lands at the next `await` (`get_db_async`, `insert_many`), and `_timer` swallows it:

```python
except asyncio.CancelledError:
    pass
```

The entries are gone from the buffer and were never written. No error is logged; the only symptom is missing request logs.

## Reproduction

Real I/O yields to the event loop, which is what delivers the cancellation, so a mock has to yield too:

```python
async def slow_insert(docs, ordered):
    await asyncio.sleep(0.01)          # real I/O yields
    inserted.extend(docs)

buffer = RequestLogBuffer(max_size=100, flush_interval=0.3)
await buffer.add(log)
await asyncio.sleep(1.5)

# inserted: 0      buffer: 0      -> the entry is lost
```

With a non-yielding mock the same test passes (2 inserts, buffer drained) — which is presumably why this has not been caught.

## Impact

Any service quiet enough not to reach 100 buffered entries loses request logs entirely. On the running stack: two cart requests never appeared; sending 120 more flushed *those* (size trigger) while the earlier two stayed missing.

The audit trail is the reason this collection exists, so silent loss on low-traffic services is worse than the latency it looks like at first.

## Suggested fix

Do not cancel the task that is doing the flushing:

```python
current = asyncio.current_task()
if self._timer_task and self._timer_task is not current and not self._timer_task.done():
    self._timer_task.cancel()
self._timer_task = None
```

Worth adding alongside: keep the entries until the write succeeds (or re-queue on failure), so a flush that fails for any reason is retried rather than dropped. And a regression test whose mocked I/O yields.

## References

- `services/commons/src/kugel_common/middleware/request_log_buffer.py` — `add` / `_reset_timer` / `_timer` / `_flush_unlocked`
- #155 — the body-size work on the same log

kugel-masa:
Closes #180.

## The defect

`RequestLogBuffer` flushes on two triggers — buffer full (100) and an idle timer (5s). **The idle one lost the batch.**

`_flush_unlocked` takes the entries out of the buffer and then cancels the timer task:

```python
to_write = self._buffer
self._buffer = []                      # entries are now only in to_write
...
if self._timer_task and not self._timer_task.done():
    self._timer_task.cancel()          # from the timer, this is the RUNNING task
```

Called from `add()` that is correct — a different task is cancelled. Called from `_timer()` itself, it cancels itself. The cancellation lands at the next `await` (inside the write) and `_timer` swallows it as `except asyncio.CancelledError: pass`. The entries were already out of the buffer, so **they were never written and nothing was logged**.

Any service quiet enough not to reach 100 buffered entries lost its request logs entirely. Hit for real while verifying #165: a single carried request did not reach MongoDB until 110 filler requests pushed the buffer past its size trigger.

## The fix

**1. Never cancel the task doing the flushing.** This is the data-loss fix.

```python
timer, self._timer_task = self._timer_task, None
if timer is not None and timer is not asyncio.current_task() and not timer.done():
    timer.cancel()
```

**2. Arm the timer once, for the oldest entry, instead of resetting it per entry.** Resetting made it an *idle* timer, which carried two costs of its own:

- a steady stream below `max_size` that never pauses for `flush_interval` kept pushing the flush out, so entries could sit indefinitely — the timer only fires when traffic stops
- it created one asyncio Task per request, which is precisely the cost this buffer exists to avoid (module docstring: "~200/sec to ~2/sec"). Measured: 20 entries created 20 tasks; now 1.

**3. A failed write no longer discards the batch silently.**

| failure | treatment | why |
|---|---|---|
| `ConnectionFailure` (backend unreachable) | kept for the next flush, bounded by `MAX_PENDING_DOCS = 1000` | a restart or failover ends; the batch is still good |
| `BulkWriteError` (server refused documents) | dropped, and the line now names how many of the batch were refused | with `ordered=False` the rest were written; repeating the batch repeats the refusal |

The backlog is a sliding window of the most recent documents and says so out loud when it drops: no bound can keep the trail complete during a long outage, so the hole is reported rather than hidden.

## About the tests

Every mocked write in the new file suspends (`await asyncio.sleep(...)`). That is load-bearing rather than decorative: **a mock that returns without suspending gives asyncio no point at which to deliver the cancellation, so the original defect passes a test written with one.** That is presumably why this was never caught.

All four behaviours mutation-checked:

|---|---|
| restore the self-cancel | 3 |
| re-arm the timer per entry | 2 |
| drop the retry backlog | 1 |
| remove the backlog bound | 1 |

The bound test initially passed the fourth mutation because it asserted against `MAX_PENDING_DOCS` itself — vacuous, since raising the constant raises the assertion with it. Rewritten to assert against the offered volume.

## Verification

Live on the rebuilt stack, one request on an otherwise idle cart service, **no filler traffic**:

```
one request sent (no fillers)          HTTP 404
waiting 8s for the idle flush (interval is 5s)...
log entries found: 1
RESULT: written by the idle flush
```

commons unit 538 · every service's unit suite · all integration · all e2e — green. `kugel_common` bumped to 0.1.74 and redistributed.

## Related

- #181 / #182 — the other two request-log defects, both still open
- #165 — the rollback trail whose value depends on this log actually being written


kugel-masa:
## Review follow-up

Reviewed by Gemini (`gemini-3.6-flash`). Three findings held up under checking and are fixed in `e040252`; a fourth turned up while writing the tests for them. Two of its findings did not survive, and two of my own assertions turned out to be vacuous.

### Fixed

**The lock was held across the database write** (raised as critical). Pre-existing — `add()` always called the flush under the lock — but wrong to leave in place now that a failed batch is retried: with an unreachable backend that is a server-selection timeout per flush, during which every concurrent `add` waits. The audit buffer would stall the requests it only means to take a note of. The write moved out of the lock; the lock now covers taking the batch and nothing else.

**A cancelled flush dropped the batch** (raised as critical). A size-triggered flush runs on the request task that happened to fill the buffer, so a client disconnect cancels it — and the batch belongs to every other request in the window, not to that client. What was not written is kept before the cancellation is allowed to stand.

**One database starved another** (raised as major). The budget was spent first-come in a loop over the targets. An outage fails every target at once, and every entry goes to both commons and the tenant database, so commons took the whole allowance and **100% of the per-tenant trail was dropped** — the copy an auditor reads. Measured, sustained outage:

```
before   commons 1000, tenant 0
after    commons  500, tenant 500     (bound holds at exactly 1000)
```

**`docs[-share:]` with a share of zero is the whole list, not an empty one.** Mine, found while mutation-testing the above. Only reachable when two flushes overlap — which is new in this commit, now that writes run outside the lock.

### Did not survive checking

**"A timer can wake, release the lock, and be seen as not-done by a concurrent `add`, leaving entries with no armed timer."** Not reachable in the committed code: `asyncio.Lock.__aexit__` calls `release()`, which does not await, so there is no suspension point between releasing the lock and the `finally` that clears the reference. It *would* have become reachable in the restructure, since the write now suspends outside the lock — so the reference is cleared in `_take_batch` under the lock instead, and there is a test that fails if it is not.

**"A retry of a partially written batch hits duplicate keys and discards the rest."** The opposite, and worth stating because it makes the retry safe: `insert_many` stamps `_id` into the documents **in place**, so a retry is idempotent. Verified against a real MongoDB:

```
after insert:  [{'a': 1, '_id': ObjectId('…')}, {'a': 2, '_id': ObjectId('…')}]
retry:         BulkWriteError, 2 refused, nInserted=0
final count:   2          ← not 4
```

### Two of my own assertions were vacuous

- the backlog bound compared `_pending_total()` against `MAX_PENDING_DOCS` itself, so it passed whatever the constant was set to — including no bound at all
- the retry test counted documents where each entry produces two (commons + tenant), so it passed on the post-outage entries alone

Both rewritten to assert against the offered volume and against the entries by URL.

### Also

Restores the module-level `get_request_log_buffer` singleton, which my first rewrite of this file dropped. Every service imports it at startup — caught by the commons suite before it left the machine.

### Verification

Eight mutations, each failing a distinct test:

|---|---|
| restore the self-cancel | 3 |
| re-arm the timer per entry | 2 |
| drop the retry backlog | 1 |
| remove the backlog bound | 1 |
| first-come retry budget | 1 |
| hold the lock across the write | 1 |
| no keep on cancellation | 2 |
| zero-share slice | 1 |
| leave the timer reference set across the write | 1 |

commons unit 546 · every service's unit suite · all integration · all e2e — green. Live single-request idle flush re-verified on the rebuilt stack.

kugel-masa:
## Second review pass

Reviewed again by Gemini (`gemini-3.6-flash`). Four findings acted on in `a8478e2`, and one of them **corrects my previous round rather than adding to it**.

### I was wrong to dismiss "in-flight writes are not awaited at shutdown"

In round 1 I argued the only background task is the timer, and shutdown cancels it. That reasoning was wrong: `_take_batch` clears `_timer_task` *before* the write begins, so a timer that got as far as writing is invisible both to `_cancel_timer` and to shutdown. Its batch died with the event loop.

Worse, the round-1 fix for the adjacent finding ("shutdown takes its batch before the cancelled timer salvages its own") was **inert for the same reason** — there is no reference left to await by then.

Mutation testing is what exposed it: the shutdown test passed with the fix reverted, which is not something reading the diff would have told me.

Tasks inside `_write` now register themselves, and shutdown **waits** for them rather than cancelling — cancelling would only move the batch into a backlog nothing is left to flush.

```python
inflight = [task for task in self._inflight if task is not asyncio.current_task()]
if inflight:
    await asyncio.gather(*inflight, return_exceptions=True)
```

### A full backlog left an empty list in `_pending`

The next flush handed it to `insert_many`. Verified against a real MongoDB:

```
insert_many([]) -> TypeError : documents must be a non-empty list
```

That failed the whole target for that flush. Empty batches are no longer stored, and `_write` skips one if it sees one anyway.

### A timer could be armed on a closing loop

`_keep_for_retry` armed one unconditionally, including from the failure path of shutdown's own final write — a task that never runs, reported by asyncio as destroyed-while-pending. Guarded by a `_closing` flag.

### Index access inside the cancellation handler

`items[attempted]` is only ever in range today, but an `IndexError` raised *there* would skip the salvage entirely. A slice costs nothing.

### Verification

Thirteen mutations now, each failing a distinct test. The two that matter most this round:

| mutation | result |
|---|---|
| shutdown does not wait for in-flight writes | `test_a_timer_flush_in_flight_is_not_abandoned` fails |
| in-flight writes are never registered | same test fails |
| arm a timer while closing | `test_shutdown_arms_no_new_timer` fails |
| keep empty lists in the backlog | `test_a_full_backlog_leaves_no_empty_list_behind` fails |
| hand an empty batch to insert_many | `test_an_empty_batch_is_skipped_rather_than_written` fails |

commons unit 550 · every service's unit suite · all integration · all e2e — green. Live single-request idle flush re-verified on the rebuilt stack.

kugel-masa:
## Third review pass

Reviewed once more by Gemini (`gemini-3.6-flash`), this time against the finished file rather than a diff. Two findings, both real, fixed in `099f019`. It reported **no** hang, deadlock, unbounded-growth or vacuous-test findings.

### A shutdown closed the buffer permanently

`_closing` was set and never cleared, so `_arm_timer` became a no-op for the rest of the process. The buffer is a module-level singleton and `shutdown()` is *also* how the integration suites force a flush — so after the first call, every entry accepted afterwards sat in the buffer with nothing scheduled to write it.

This one is mine, introduced by the round-2 zombie-timer guard. The flag is now held only for the duration of the call, which is all that guard ever needed.

### A wrapped connection failure was treated as permanent

`get_db_async` wraps everything it catches in a `DatabaseException`, so an unreachable backend can reach the handler with the `ConnectionFailure` only in `__cause__` — and `except ConnectionFailure` would miss it, turning an outage into permanent loss instead of a retry. Classification now walks the cause chain.

Worth being precise about how bad this was: in the ordinary path the failure surfaces at `insert_many`, not at `get_db_async`. Verified against a dead port:

```
client + db handle obtained without error: True
insert_many raised ServerSelectionTimeoutError; ConnectionFailure? True
```

So the retry did work — by luck rather than by design.

Everything that is *not* a connection failure stays permanent on purpose: a document the driver refuses to encode, one over the BSON size limit, a duplicate key. Repeating those would hold the backlog against entries that could still be written. The log line now says the entries are lost, rather than only that the flush failed.

### Not addressed

Two concurrent `shutdown()` calls, where the second can return while the first is still writing. The lifespan calls it once and the suites call it sequentially; guarding it would add state for a case that does not occur. Recorded here rather than fixed.

### Verification

Seventeen mutations across three rounds, each failing a distinct test. This round:

| mutation | test that fails |
|---|---|
| closing flag never reset | `test_the_buffer_still_works_after_a_shutdown` |
| classify only the exception in hand | `test_a_connection_failure_wrapped_by_the_db_helper_is_still_retried` |
| retry every failure, transient or not | `test_a_failure_a_repeat_cannot_help_is_not_retried` |
| no timer guard while closing | `test_the_final_write_arms_no_timer` |

commons unit 553 · every service's unit suite · all integration · all e2e — green. Live single-request idle flush re-verified on the rebuilt stack.

### Where the three rounds landed

Round 1 found four real defects, round 2 four (one of which corrected round 1's own fix), round 3 two — one of which was introduced by round 2. The rate is falling and the last round found nothing in the areas it was pointed at hardest (hangs, growth, test vacuity), which is the signal I was waiting for.

Issue #161, opened 2026-08-13
commons: request-logging middleware turns a 401 into a 500 and drops the request log


kugel-masa:
Found while reviewing #160.

## The defect

`RequestLoggingMiddleware` calls `_get_terminal_info` as the first statement of its
`finally` block (`services/commons/src/kugel_common/middleware/log_requests.py:76`):

```python
        finally:
            is_terminal_service = True if service_name == "terminal" else False
            terminal_info = await _get_terminal_info(request, is_terminal_service)
            user_dict = await _get_current_user(request)
            request_log = RequestLog(...)
```

`_get_terminal_info` (`:199-231`) calls `kugel_common.security.get_terminal_info`
without guarding it. That function maps **any** `HttpClientError` onto
`HTTPException(401, "Invalid API key")` (`security.py:316-328`), so the exception
escapes the `finally` block.

The sibling helper on the very next line is already hardened against exactly this,
and its docstring says why (`:234-241`):

> This helper is used from the request-logging middleware's `finally` block — if
> `get_current_user` raises (invalid / expired / forged JWT) we MUST NOT let that
> escape, otherwise it suppresses the original auth failure being propagated by
> the route handler and FastAPI ends up returning 500 instead of the proper 401
> to the client.

`_get_terminal_info` runs *before* `_get_current_user` and has no such guard.

## Consequences

An exception from the terminal lookup does three things at once:

1. **Replaces the route's response.** Whatever the handler produced — success or a
   correctly-formed error — is discarded and the client gets a 500.
2. **Drops the request log.** The rest of the `finally` block never runs, so no
   `RequestLog` is built, written to file, or buffered for the database. Logging is
   lost precisely for the requests that failed.
3. **Adds latency** when the cause is a connection failure: `HttpClientHelper`
   retries three times with a one-second delay before raising.

## Trigger

`_get_terminal_info` fires whenever a request carries an `X-API-KEY` header
together with a `terminal_id`, which may come from a query parameter, so it is
client-controlled and reachable on any route of any service (`:216-226`).

The lookup then fails for any of:

- the terminal service being unreachable or slow
- `BASE_URL_TERMINAL` not being configured (see #159)
- **a genuinely invalid API key**, since the terminal service answers 401 and that
  becomes an `HttpClientError` too

### An invalid API key answers 500 instead of 401, on every service

The clearest form needs no misconfiguration at all. Same bad key, same route; the
[...]

Issue #67, opened 2026-03-23
feat: Implement lifecycle-aware JWT token authentication for terminal API key


kugel-masa:
## Summary

Replace the current API key verification flow (per-request calls to terminal service) with a JWT token-based authentication system. Tokens are issued at terminal lifecycle state changes, enabling local verification at each service without inter-service calls.

### Current Flow (Problem)
```
POS Terminal → [X-API-KEY] → Each Service → [HTTP call] → Terminal Service → DB lookup
```
- Every request from every service triggers an HTTP call to the terminal service for API key validation
- Cart service additionally uses `api_key` from `terminal_info` to call master-data service
- High inter-service traffic and latency for authentication

### Proposed Flow
```
1. POS Terminal → [X-API-KEY] → POST /auth/token → JWT issued
2. POS Terminal → [JWT] → Any Service → Local JWT verification (no inter-service call)
```

## Design: Lifecycle-Aware Token

Terminal state changes only occur at well-defined lifecycle operations (open, sign-in, sign-out, close). The JWT is re-issued at each of these state changes, ensuring token claims always reflect current state.

### JWT Claims
```json
{
    "tenant_id": "tenant001",
    "store_code": "001",
    "terminal_no": "01",
    "terminal_id": "tenant001-001-01",
    "staff_id": "S001",
    "staff_name": "Tanaka Taro",
    "business_date": "20260323",
    "open_counter": 1,
    "business_counter": 5,
    "status": "Opened",
    "iss": "terminal-service",
    "token_type": "terminal",
    "iat": 1742691600,
    "exp": 1742778000
}
```

### Token Issuance / Refresh Timing

| API | Operation | Token |
|-----|-----------|-------|
| `POST /auth/token` | Initial auth (API key → JWT) | **New** |
| `POST /terminals/{id}/open` | Terminal open | **Re-issue** (business_date, counters) |
| `POST /terminals/{id}/sign-in` | Staff sign-in | **Re-issue** (staff added) |
| `POST /terminals/{id}/sign-out` | Staff sign-out | **Re-issue** (staff removed) |
| `POST /terminals/{id}/close` | Terminal close | **Re-issue** (status updated) |
| Other services (cart, master-data, etc.) | Business operations | No update (local verification only) |

### Token Delivery
- New tokens are returned via `X-New-Token` response header on state-change APIs
- Client stores the new token and uses it for subsequent requests
- No impact on response body schemas

### Token Expiry & Refresh
[...]

PR #184, opened 2026-08-21
fix(commons): record the store and terminal for a JWT-authenticated request (#181)

### LINKED ISSUE #181: commons: a terminal-JWT request logs no store or terminal, so the request log loses attribution as clients migrate

kugel-masa:
Closes #181.

## The defect

`_get_terminal_info` resolves the terminal from `X-API-Key` plus a `terminal_id` parameter. A terminal-JWT request carries **neither** — that is the point of the migration (`docs/ja/terminal/frontend-jwt-migration-guide.md`: 「cart は `terminal_id` クエリも `X-API-KEY` も不要になります」).

So the request log recorded an empty terminal for exactly the credential the fleet is moving to. Measured, same request, both credentials:

```
API key  terminal_info={"store_code":"5678","terminal_no":9,"business_date":"20260821","open_counter":1}
         staff_info={"id":"S001","name":"Staff1"}

JWT      terminal_info={"store_code":"","terminal_no":0,"business_date":"","open_counter":0}
         staff_info={"id":"","name":""}
```

Store, terminal number, business date, open counter and the signed-in staff were all lost. The identity survived only as text inside `user_info.username` (`"terminal:T6216-5678-9"`), so queries by store, terminal or business date stop working as clients migrate — and a request cannot be tied to an open session at all.

## The fix

The claims already carry every one of those fields, and `terminal_claims_to_terminal_info` already builds a `TerminalInfoDocument` from them. So this reads rather than looks up — no database round trip on a path that runs for every request. The API-key path keeps precedence where both are present.

After, on the running stack:

```
JWT      terminal_info={"store_code":"5678","terminal_no":9,"business_date":"20260821","open_counter":1}
         staff_info={"id":"S001","name":"Staff1"}
```

## The part that needed the most care

The helper returns `None` and **never raises**. It is called from the logging middleware's `finally` block, where an exception replaces whatever the route was returning — including the 401 a bad token is supposed to produce. That is the defect class of #161, and it would be a poor trade to fix attribution by introducing it.

Four tests hold that line: no token, a garbage token, a token signed with the wrong key, and a route that fails on its own terms while the token is also bad (the response must still be the route's 500, and the log must still be written).

## Verification

|---|---|
| revert the JWT fallback | 4 |
| let the token decode raise | 4 |

commons unit 567 · every service's unit suite · all integration · all e2e — green. `kugel_common` bumped to 0.1.75 and redistributed; 0.1.74 is already on main as the #180 buffer fix, so this needs its own version rather than redefining that one.

## Related

**This is a prerequisite for #182.** Fixing that index while JWT requests still recorded empty values would enforce per-terminal uniqueness on `""`/`0` for every JWT request — recreating the same tenant-wide key space for exactly the traffic the fleet is migrating to.

- #182 — the unique index keyed on fields that do not exist
- #161 — the `finally`-block defect class this fix had to avoid
- #67 / #68 — the terminal JWT work this follows from

### DIFF
--- a/services/account/Pipfile
+++ b/services/account/Pipfile
@@ -13,7 +13,7 @@ motor = "*"
 pytz = "*"
 pyjwt = "*"
 passlib = {extras = ["bcrypt"], version = "*"}
-kugel_common = {file = "commons/dist/kugel_common-0.1.74-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.75-py3-none-any.whl"}
 
 bcrypt = "==3.2.0"
 python-multipart = "*"
--- a/services/account/Pipfile.lock
+++ b/services/account/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "2deaad3a04b72da49c6bdda3db0ccaaa42ec70cc167a95778c8a167c5ffae28a"
+            "sha256": "9ea556962fff9c8a6628c90c42afbbfa672dea411c4b8306c281a7979516a6df"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -667,9 +667,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.74-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.75-py3-none-any.whl",
             "hashes": [
-                "sha256:18ac251853bdd781eebbd1d9fad7985ce40bf71ee007741ab907fa5305939588"
+                "sha256:356e3ea3732e5c7ddb2f254a2b999a2f9ddcd71548126b7e853a85ad7e1bb980"
             ]
         },
         "motor": {
--- a/services/cart/Pipfile
+++ b/services/cart/Pipfile
@@ -12,7 +12,7 @@ pydantic-settings = "*"
 motor = "*"
 pytz = "*"
 pyjwt = "*"
-kugel_common = {file = "commons/dist/kugel_common-0.1.74-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.75-py3-none-any.whl"}
 httpx = "*"
 aiohttp = "*"
 wcwidth = "*"
--- a/services/cart/Pipfile.lock
+++ b/services/cart/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "27efbb484e0b3f18cb116330264f6e74e6a377acc621186db580ee59e43246ed"
+            "sha256": "0e6bfc2b4edb5c79dee8c065ef3bd99b95a7a90b3f4eeb0b2efac919f192e2b3"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -659,9 +659,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.74-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.75-py3-none-any.whl",
             "hashes": [
-                "sha256:18ac251853bdd781eebbd1d9fad7985ce40bf71ee007741ab907fa5305939588"
+                "sha256:356e3ea3732e5c7ddb2f254a2b999a2f9ddcd71548126b7e853a85ad7e1bb980"
             ]
         },
         "motor": {
--- a/services/cart/tests/integration/test_request_log_jwt_attribution.py
+++ b/services/cart/tests/integration/test_request_log_jwt_attribution.py
@@ -1,0 +1,76 @@
+# Copyright 2026 masa@kugel
+"""What a terminal-JWT request leaves in the request log (issue #181).
+
+Here rather than in `services/commons` because that package is unit-only by
+design, and what matters is the row that reaches MongoDB. The commons tests prove
+the middleware builds the right document; this proves it is stored, through a
+real service and a real collection.
+
+The token used is the suite's existing `terminal_jwt` fixture, which carries no
+`open_counter` claim - the shape a terminal has before it is ever opened. That
+was not deliberate on its part, but it is the case worth running against a real
+stack: an absent claim reaching a required field raises inside the middleware's
+`finally`, where the cost is not a missing field but the route's response.
+"""
+
+import os
+
+import pytest
+from kugel_common.middleware.request_log_buffer import get_request_log_buffer
+
+pytestmark = pytest.mark.asyncio
+
+
+async def _logged(url_fragment):
+    """Flush the request-log buffer, then read back the row for this request."""
+    from kugel_common.database import database as db_helper
+
+    await get_request_log_buffer().shutdown()
+    db = await db_helper.get_db_async(f"{os.environ.get('DB_NAME_PREFIX')}_{os.environ.get('TENANT_ID')}")
+    return await db["log_request"].find_one(
+        {"request_info.url": {"$regex": url_fragment}}, sort=[("request_info.accept_time", -1)]
+    )
+
+
+async def test_a_jwt_request_records_the_store_and_terminal(http_client, terminal_jwt):
+    """Before #181 this row named no store, no terminal and no staff.
+
+    The request itself is refused - the fixture's terminal is Idle, not opened -
+    which is the point: a refusal is what an audit trail most needs to attribute,
+    and it is refused on the route's own terms rather than by the logger.
+    """
+    response = await http_client.post(
+        "/api/v1/carts",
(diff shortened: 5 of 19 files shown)


kugel-masa:
## Review follow-up

Reviewed by Gemini (`gemini-3.6-flash`). Two findings acted on in `5172aef` — **one of them a regression this PR introduced**, and a serious one.

### A token from a never-opened terminal turned a 200 into a 500

`create_terminal_token` omits the claims for state the terminal does not have yet, so `open_counter` arrives as `None` — and `RequestLog.TerminalInfo` declares it `int`. Before this PR that never mattered, because a JWT request never populated `terminal_info` at all. Now it does, and the `ValidationError` escapes the middleware's `finally`: the route's response is replaced and the log is dropped. Reproduced:

```
open_counter on a never-opened terminal: None
response status: 500   (200 expected; 500 = the finally hijacked it)
logs recorded: 0
```

Every field is now defaulted, not only `business_date`.

**Worth stating why no suite caught it.** Every terminal an existing tenant has was opened long ago, so `open_counter` is set on every token the tests ever see. It takes a terminal's *first* token — which is exactly what a new deployment starts with. Full unit + integration + e2e were green over this defect.

This is the defect class of #161. That issue is about the **API-key** branch reaching it and stays open; what is fixed here is the new way in that this PR opened.

### The API key took precedence over the token

The routes resolve the other way round — `get_terminal_info_with_jwt_or_apikey`: *"Priority 1: Try terminal JWT"*. So a request carrying both **executes** as the token's terminal while the log **named** the API key's. Beyond the wrong attribution, the API-key branch costs an HTTP call to the terminal service that a token-authenticated request has no reason to pay. The order now matches the routes.

I had reached the same conclusion independently while the review was running, by reading the route dependency.

### Findings not acted on

- **The `finally` block is unprotected in general.** Correct, and it is #161 — filed, open, and next in the queue. Guarding the whole block here would close that issue as a side effect, without the test that proves a 401 survives, which is the specific thing #161 is about.
- **The token is decoded twice per request.** True (`_get_terminal_info` and `_get_current_user` use different verifiers with different claim shapes). An HS256 verify of a small token is tens of microseconds; sharing the decode would mean restructuring both helpers around a common cache. Not worth it here.
- **`test_the_tenant_is_still_recorded` is vacuous.** Correct — `_make_tenant_id` falls back to the decoded user, so it passes with the fix reverted. Kept as a guard that the new path does not break the tenant, and now says so rather than being offered as evidence.

### Verification

|---|---|
| revert the JWT fallback | 4 |
| let the token decode raise | 4 |
| stop defaulting `open_counter` | 1 |
| restore API-key precedence | 1 |

commons unit 570 · every service's unit suite · all integration · all e2e — green. Live attribution re-verified on the rebuilt stack.

kugel-masa:
## Second review pass — no findings

Reviewed again by Gemini (`gemini-3.6-flash`) against the corrected state. It reported **no critical, major or minor findings**, and walked the specific questions it was pointed at: the flipped order, every claim's type, the falsy-default question, the spoofing surface, and test vacuity.

One thing it reasoned about was not actually covered by any test, so I added it (`1b75d85`).

### A validly signed token with wrong-shaped claims

The review argued this is safe because the document is constructed *inside* the guard. That is correct — and nothing was pinning it. A signature proves the claims were issued unmodified, not that they are sane, and this runs in the `finally` where a `ValidationError` does not merely lose the attribution: it replaces the route's response.

Six signed tokens now assert the route still answers and the request is still logged:

```
terminal_no is a word · store_code is a list · open_counter is an object
business_date is a number · staff_id is a number · explicit nulls
```

Mutation-checked: moving the document construction outside the guard fails all six.

### Where this landed

Round 1 found two real defects — one of them a regression this PR introduced, which turned a 200 into a 500 for a terminal's first token and which the full suite ran straight past. Round 2 found none.

|---|---|
| revert the JWT fallback | 4 |
| let the token decode raise | 4 |
| stop defaulting `open_counter` | 1 |
| restore API-key precedence | 1 |
| construct the document outside the guard | 6 |

commons unit 576 · every service's unit suite · all integration · all e2e — green. Live attribution verified on the rebuilt stack.

kugel-masa:
## Coverage audit — and a correction

### Correction first

An earlier comment on this PR said a terminal's **first token** turns a 200 into a 500. **That is wrong.** Measured against the running stack:

```
create terminal_no=90: HTTP 201
stored openCounter: 0   businessDate: None
claims present: open_counter=True
GET with that first token: HTTP 200
```

A terminal created through the terminal service stores `open_counter` as **0, not None**, so the claim is present and the path is not reachable that way. Across 433 stored terminals the only document without one is an empty `{}` with no `terminal_id`, which can never issue a token.

So the suites did **not** run past a live defect — they were right that the API path is safe. What is real is narrower, and still worth the fix: **any token whose claims omit `open_counter`** — a forged one, or one minted from a document that does not set the field — hijacks the response rather than merely losing a field.

### The coverage gap

Everything so far was commons unit tests over a captured buffer. That proves the middleware builds the right document; it does not prove the row reaches the collection. The only evidence for that was a manual probe — the same gap #183 had.

Two integration tests added in cart, using the suite's **existing `terminal_jwt` fixture**. That fixture already existed and **no test used it**, which is why the JWT path had no integration coverage at all.

It also carries no `open_counter` claim, which makes it the exact shape that raises. So it doubles as the reachable-in-practice case, against a real service and a real collection:

| mutation | result |
|---|---|
| stop defaulting `open_counter` | both integration tests fail |
| revert the JWT fallback | the attribution test fails |

The first test deliberately asserts on a request that is **refused** (the fixture's terminal is Idle, not opened): a refusal is what an audit trail most needs to attribute, and it must be refused on the route's terms rather than by the logger.

### One flaky failure, recorded

A full e2e run failed once on `journal/test_setup_data` — `POST /tenants` returned 500 with "Error creating tenant: T6216". It did not reproduce on the journal suite alone or on a repeat full run (both green). Not from this change as far as I can tell, but recording it rather than dropping it.

### Totals

commons unit 576 · cart integration 60 · every service's unit suite · all integration · all e2e — green.

Issue #185, opened 2026-08-21
commons: tenant setup fails permanently, and says nothing, when a collection already holds data that violates one of its unique indexes


kugel-masa:
Found while investigating a flaky e2e failure: `journal/test_setup_data` returned 500 once during a full run and did not reproduce on a rerun.

```
AssertionError: {"success":false,"code":500,"message":"Error creating tenant: T6216", ...}
assert 500 in (201, 400)
```

## The mechanism

`POST /tenants` creates each collection and then ensures its indexes. The indexes come **after** the collection exists, so anything already written to that collection has to satisfy them. If it does not, the index cannot be built and the whole tenant setup fails.

Reproduced deterministically. One open/close event, delivered twice into a database whose unique index does not exist yet:

```
duplicate open/close present -> POST /tenants: HTTP 500
single event present        -> POST /tenants: HTTP 201
retries after the failure   : [500, 500, 500]
```

The service log names what actually happened; the API response does not:

```
Failed to execute command: {'createIndexes': 'log_open_close', 'indexes': [...], 'unique': True}
Required index (...) missing on log_open_close after migration
  (index ensure failed — likely existing data violates a new unique constraint)
Error creating tenant: T6216
```

## Why data can be there first

The window is structural, not exotic:

- the e2e conftests drop the tenant database at session start, and the service keeps consuming pub/sub while `test_setup_data` runs
- a fresh deployment has no database at all, so any event that arrives before setup creates its collection
- restoring or clearing a tenant database has the same effect

And in that window the collection's own duplicate protection is gone. `OpenCloseLogRepository.create_open_close_log` guards with a check-then-act:

```python
if await self.get_one_async(filter=filter):
    logger.warning("OpenCloseLog already exists...")
    return open_close_log
...
await self.create_async(open_close_log)
```

That is only safe because a unique index normally backs it. Before the index exists, two at-least-once deliveries of the same event can both pass the check and both insert — and then the index can never be built.

## Why it matters

- **It does not heal.** Every retry returns the same 500 until someone finds and removes the offending documents by hand. A tenant in this state cannot be set up at all.
- **The response says nothing.** `app/api/v1/tenant.py` catches `Exception` and logs only `Error creating tenant: {tenant_id}`. The collection, the index and the reason exist in the service log and nowhere else, so an operator (or a failing CI run) is handed a bare 500.
[...]

PR #186, opened 2026-08-22
fix(commons): say what blocks a tenant setup, and finish the collections that can be (#185)

kugel-masa:
Refs #185.

## The defect

A tenant setup that hit a collection whose unique index could not be built returned this and nothing else:

```
{"success":false,"code":500,"message":"Error creating tenant: T6216", ...}
```

No collection, no index, no reason — and the same 500 on every retry. Everything needed to act on it existed one frame down and was discarded.

Observed on real data during this work: a journal held **transaction 1 four times**, so `log_tran`'s unique index could never be rebuilt.

## Three changes, one gap

**Name the documents that are in the way.** The message said the index was *"likely"* blocked by existing data. Now it looks:

```
Required index (('tenant_id', 1), ('store_code', 1), ('terminal_no', 1),
('business_date', 1), ('open_counter', 1), ('operation', 1)) missing on
log_open_close after migration (index ensure failed). Documents already in the
collection share this key and have to be resolved before it can be built
(showing up to 5): {'tenant_id': 'T6216', ..., 'open_counter': 7,
'operation': 'open'} x3
```

`find_blocking_duplicates_async` honours a partial filter — documents outside it are not indexed and so cannot be what stops the build; naming them would send an operator after the wrong rows. It is best-effort: an aggregation that fails must not replace the failure it was called to explain.

**Finish the collections that can be finished.** `create_collections` stopped at the first failure, so one blocked collection also left every collection *after* it uncreated — and the operator learned about the blocked ones one restart at a time. All seven services now run their steps through `run_setup_steps_async`, which attempts every step and reports all the failures at once. An unreachable backend is still different in kind and stops immediately, since the remaining steps could only fail the same way.

**Carry the reason into the response.** The tenant endpoints discarded it, and terminal's fan-out used `raise_for_status()` — which reports the URL and the status and drops the body, the body being where the downstream service names the collection. That fan-out is the point an operator meets first; it is what produced the original `Server error '500 Internal Server Error' for url 'http://journal:8000/api/v1/tenants'`.

**Also**: the end-state is now verified on newly created collections, not only migrated ones. The ensure loop swallows its failures with a warning, so an index that never got built on a new collection reported success.

## Verified on the running stack

Three duplicate open/close events, then `POST /tenants`:

```
HTTP 500
names the collection : True
names the duplicates : True
names the key values : True

500: Error creating tenant: T6216: Tenant setup for T6216 could not complete
1 of 5 steps: Failed to create collection with indexes: log_open_close in ...
```

Note `1 of 5` — the other four collections were set up.

|---|---|
| do not name the blocking documents | e2e + integration |
| stop at the first failing step | e2e + unit |

Both run against the **running container**, not the source tree, since an e2e that cannot fail is worth nothing.

commons unit 581 · every service's unit suite · all integration · all e2e — green. `kugel_common` bumped to 0.1.76.

## Not addressed, and left on the issue

**How the duplicates get in.** The repositories guard with a check-then-act that is only safe while the unique index exists — and the window this issue is about is precisely when it does not. Two at-least-once deliveries can both pass the check and both insert. Closing that is a change to the write paths across services, and a different piece of work from making the failure legible.

So this PR makes a blocked tenant diagnosable and repairable; it does not stop a tenant becoming blocked.

### DIFF
--- a/services/account/Pipfile
+++ b/services/account/Pipfile
@@ -13,7 +13,7 @@ motor = "*"
 pytz = "*"
 pyjwt = "*"
 passlib = {extras = ["bcrypt"], version = "*"}
-kugel_common = {file = "commons/dist/kugel_common-0.1.75-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.76-py3-none-any.whl"}
 
 bcrypt = "==3.2.0"
 python-multipart = "*"
--- a/services/account/Pipfile.lock
+++ b/services/account/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "9ea556962fff9c8a6628c90c42afbbfa672dea411c4b8306c281a7979516a6df"
+            "sha256": "8c7a4e6e8d9f206486c8504b0fc41c17ab873aff5b63af9c3a5b165b64a1da76"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -667,9 +667,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.75-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.76-py3-none-any.whl",
             "hashes": [
-                "sha256:356e3ea3732e5c7ddb2f254a2b999a2f9ddcd71548126b7e853a85ad7e1bb980"
+                "sha256:480292c99322b0fe4bcb9eda5332aad2de09c29a58c19dddfc4b36b1e7d1797a"
             ]
         },
         "motor": {
--- a/services/account/app/database/database_setup.py
+++ b/services/account/app/database/database_setup.py
@@ -81,21 +81,19 @@ async def create_request_log_collection(tenant_id: str):
 
 
 async def create_collections(tenant_id: str):
+    """Create every collection this service needs for a tenant.
+
+    Through run_setup_steps_async so that one blocked collection does not stop
+    the rest from being created, and so a failure names every blocked collection
+    at once rather than one restart at a time (issue #185).
     """
-    Create all required collections for a new tenant
-
-    This function should be extended when new collections are added to the system.
-    Each collection creation should be added to this function.
-
-    Args:
-        tenant_id: The tenant identifier
-
-    Returns:
-        None
-    """
-    await create_user_account_collection(tenant_id)
-    await create_request_log_collection(tenant_id)
-    # Add more collections here as the system grows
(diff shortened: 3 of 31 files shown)


kugel-masa:
## Review follow-up

Reviewed by Gemini (`gemini-3.6-flash`). Four findings acted on in `795191c`, two declined — and one of the fixes turned out to introduce a defect of its own.

### The diagnostic aggregation was unbounded

It runs on a collection with no usable index for the grouping — a full scan — and the collections it runs on are production logs that can hold millions of documents. Now `allowDiskUse` with a 10s ceiling. A failure without the sample beats spending the database to explain one, and timing out degrades to the empty list it already handled.

### The fan-out echoed the downstream body whole

What answers is not always one of our services saying something short; a proxy in between returns an HTML error page. Truncated to 1000 characters, and it says when it cut.

### A unique index that is not unique read as satisfied

The interesting one. Checked against the running server rather than assumed:

```
after non-unique create: [["_id_",false],["already_here",false]]
createIndexes(unique, new name) -> ok=1
final: [["_id_",false],["already_here",false],["probe_unique",true]]
```

MongoDB 7.0 lets both coexist under different names. So when the unique build **fails** and a non-unique index on the same keys exists, the keys are present, the keys-only check passes, and a constraint the system relies on is reported as there while nothing enforces it. Now caught, and it names the documents keeping it non-unique.

### The defect that fix introduced

The first version kept one index per key pattern in a dict — so with both present it decided by **listing order**, and could fail a collection whose constraint is present and enforced. Grouped instead.

Mutation testing is what found it: collapsing the grouping back into a dict failed nothing, because the ordering case had no test. It does now.

### Declined

**`OperationFailure` should stop the setup the way a connection failure does.** It also covers per-collection failures — an index build conflict *is* an `OperationFailure` — so stopping on it would abort a setup that has one bad collection and six good ones. That is the thing this PR set out to fix.

**The fan-out error propagation is untested.** It is; the test lives in `services/terminal/tests/`, which I left out of the diff I sent for review. My omission, not a gap.

### Documented rather than fixed

`find_blocking_duplicates_async` does not resolve a unique index on an **array** field: `$group` groups on the array as a value while the index collides on the elements, so `[1, 2]` and `[2, 3]` read as distinct here and clash in the index. No unique index in this system is on an array field; the docstring now says what would need to change if one is added.

### Verification

|---|---|
| do not name the blocking documents | 2 |
| stop at the first failing step | 2 |
| drop the index-options check | 1 |
| collapse the grouped indexes into a dict | 1 |
| ignore the partial filter when sampling | 1 |
| echo the downstream body whole | 1 |

commons unit 581 · every service's unit suite · all integration · all e2e — green.

PR #187, opened 2026-08-22
fix: key the request-log index on the fields a request log actually has (#182)

### LINKED ISSUE #182: commons: the request-log unique index keys on fields that do not exist, so it constrains a tenant-wide timestamp instead of a terminal

kugel-masa:
Closes #182.

## The defect

Every service created its `log_request` collection with this index:

```python
{"keys": {"tenant_id": 1, "store_code": 1, "terminal_no": 1, "request_info.accept_time": 1},
 "unique": True}
```

A request log document has **no `store_code` or `terminal_no` at the top level** — they live under `terminal_info`:

```
_id, shard_key, created_at, updated_at, etag, tenant_id, client_info,
request_info, response_info, staff_info, user_info, terminal_info, snapshot_info
```

Missing fields index as null, so the key was `(tenant_id, null, null, accept_time)` for every document. The constraint it actually enforced was **one request per tenant per timestamp**, across every terminal in that tenant. It has never done the job it was written for, under any credential.

## Two changes

**The paths are corrected** to `terminal_info.store_code` / `terminal_info.terminal_no`.

**The uniqueness is dropped**, and that is the deliberate part.

What a unique index here could protect against — the same request written twice — does not happen. `insert_many` stamps `_id` into the documents in place, so a retried batch is refused by `_id`, verified in #183:

```
after insert:  [{'a': 1, '_id': ObjectId('…')}, {'a': 2, '_id': ObjectId('…')}]
retry:         BulkWriteError, 2 refused, nInserted=0
final count:   2          ← not 4
```

What it *did* do was reject audit records that legitimately happened, and drop them silently. Measured by comparing the tenant databases against the commons copies, which carry no such index:

```
audit rows the current constraint would have dropped: 66
```

An audit trail should not be the thing that decides a record did not happen. Kept as a query index, which is what it is genuinely useful for.

## Migration

The stale index is retired by key pattern through the existing `drop_indexes_by_keys` mechanism, and the end-state verification added in #186 hard-fails if it survives. Three services' `create_some_collection` did not forward that parameter and now do.

Checked against the running stack, on a collection put back into the pre-#182 state:

```
before: [["tenant_id,store_code,terminal_no,request_info.accept_time", true]]
after : [["tenant_id,terminal_info.store_code,terminal_info.terminal_no,request_info.accept_time", false]]
```

**Existing data cannot block it.** Scanned 259 tenant `log_request` collections holding 21,929 rows: **0** collide under the corrected key. That matters because #185 is exactly what happens when they do.

## This needed #181 first

Before it, JWT requests recorded `terminal_no: 0`. The corrected index would then have started enforcing per-terminal uniqueness over rows that all claimed the same terminal — and one of the 66 dropped rows is precisely that case:

```
db_cloud_cart_commons {"ts":"2026-08-08T09:38:34.495808+09:00","rows":2,"terminals":[0]}
```

## Verification

|---|---|
| restore the wrong key paths | all 5 (setup can no longer complete) |
| restore `unique` | 2 |
| skip the migration drop | 1 |

commons unit 581 · every service's unit suite · all integration · all e2e — green.

## Related

- #181 — the prerequisite, merged
- #185 — what a blocked index costs, and the reporting that would now name it
- #183 — where the `_id` idempotency was established

### DIFF
--- a/services/account/app/database/database_setup.py
+++ b/services/account/app/database/database_setup.py
@@ -19,7 +19,9 @@ from app.config.settings import settings
 logger = getLogger(__name__)
 
 
-async def create_some_collection(tenant_id: str, collection_name: str, index_keys_list: list, index_name: str):
+async def create_some_collection(
+    tenant_id: str, collection_name: str, index_keys_list: list, index_name: str, drop_indexes_by_keys: list = None
+):
     """
     Create a collection with specified indexes in the tenant's database
 
@@ -28,13 +30,20 @@ async def create_some_collection(tenant_id: str, collection_name: str, index_key
         collection_name: Name of the collection to create
         index_keys_list: List of index configurations to create on the collection
         index_name: Base name for the indexes
+        drop_indexes_by_keys: Index key patterns to retire first, matched by
+            pattern rather than by name (issue #182 uses it to replace the
+            index keyed on fields a request log document does not have)
 
     Returns:
         None
     """
     db_name = f"{settings.DB_NAME_PREFIX}_{tenant_id}"
     await db_helper.create_collection_with_indexes_async(
-        db_name=db_name, collection_name=collection_name, index_keys_list=index_keys_list, index_name=index_name
+        db_name=db_name,
+        collection_name=collection_name,
+        index_keys_list=index_keys_list,
+        index_name=index_name,
+        drop_indexes_by_keys=drop_indexes_by_keys,
     )
 
 
@@ -62,8 +71,9 @@ async def create_request_log_collection(tenant_id: str):
     """
     Create the request log collection with appropriate indexes
 
-    Creates a collection for storing API request logs with a compound index
-    on tenant_id, store_code, terminal_no, and request acceptance time.
+    Creates a collection for storing API request logs, indexed by tenant,
+    store, terminal and request acceptance time (issue #182: by the paths those
+    last two actually live at, and no longer unique).
 
     Args:
         tenant_id: The tenant identifier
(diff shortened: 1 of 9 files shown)


kugel-masa:
## Review

Reviewed by Gemini (`gemini-3.6-flash`): **no findings** — no critical, major or minor. It worked through the specific questions it was pointed at: what could depend on the uniqueness, whether dropping by key pattern can take the wrong index, whether the new key order suits how the collection is read, MongoDB's handling of the option change, the changed helper signatures, and test vacuity.

The one answer worth not taking on trust is the first, so I checked it myself rather than accept it.

### Nothing depends on the uniqueness — verified

`log_request` has exactly two writers and no readers in this repository:

- `RequestLogBuffer` (`insert_many`, batched) — the live path
- `RequestLogRepository.create_request_log_async` — a plain insert, no upsert, no dedupe

And the second turns out to be **dead code**. `_output_request_log_to_db_async` and `_output_request_log_to_db` have no call sites; the buffer replaced that path and the functions were left behind. So the only live writer is the buffered `insert_many`, whose idempotency comes from `_id`, not from this index.

No repository, report, or query reads the collection at all. There is nothing to break.

(The dead path is worth removing, but not here — it sits in the same file #161 will touch.)

### Two of its answers are worth recording

**On the key order.** `tenant_id → store_code → terminal_no → accept_time` puts the equality fields first and the range/sort field last, which is the order a query filtering by tenant/store/terminal and ranging on time wants. That was not a deliberate design choice on my part — the corrected index inherits the order the broken one had — but it happens to be the right one.

**On the option change.** MongoDB will not alter `unique` in place, so this only works because the key paths changed too: it is a new index rather than an option change on an existing one, and the stale one is dropped before the new one is ensured. Had the paths been right and only the uniqueness wrong, the migration would have needed a different shape.

### State

commons unit 581 · every service's unit suite · all integration · all e2e — green. Migration verified against the running stack on a collection put back into the pre-#182 state.

PR #188, opened 2026-08-22
fix(commons): stop the request logger from deciding the response (#161)

### LINKED ISSUE #161: commons: request-logging middleware turns a 401 into a 500 and drops the request log

kugel-masa:
Closes #161.

## The defect

The middleware assembles its record in a `finally` block. That block runs while the route's own exception is still in flight — so **an exception raised there replaces it**.

`_get_terminal_info` called `security.get_terminal_info` unguarded, and that maps **any** `HttpClientError` onto `HTTPException(401)`. A terminal service that was merely unreachable therefore arrived as an authentication failure, escaped the block, and turned a request the route had already answered into a 500 with no request log at all.

Reported from a live stack as a **successful** authentication answered 500, three seconds late, with a false audit entry and nothing logged.

## Four guards, because they buy different things

**The terminal lookup degrades to `None`.** The request is then still recorded with everything else that is known — where letting the exception out records nothing.

**The whole `finally` body is wrapped.** Not only the terminal lookup: every call in there is observability, and none of it gets a vote on the response. This is the one that holds if the buffer, the file logger, or anything added later fails.

**Capturing the request is wrapped too.** `_make_request_info` runs *before* the route, so a failure reading or sanitizing the body used to answer the caller with the logger's error before the route was ever called. Not part of the original report; found while writing the tests.

**The record falls back to a minimal `request_info`** when that capture failed, so the request is logged rather than lost to a validation error. The failures are what most need a trail.

## Also: a dead write path removed

`_output_request_log_to_db` and `_output_request_log_to_db_async` have no call sites — the buffer replaced that path and the functions were left behind. A second, unreferenced write path in an audit component is a trap for the next reader. Found while checking what depends on the request log for #182.

## Verified on the running stack

With an API key the terminal service refuses, which is the shape that made the middleware's own lookup raise:

```
request with a refused API key -> HTTP 401     (was 500)
request logged: True
recorded status: 401
```

## Verification

|---|---|
| unguard the terminal lookup | 2 |
| remove the guard around the `finally` body | 1 |
| unguard the request capture | 1 |
| drop the `request_info` fallback | 1 |

Worth noting what the first one shows: unguarding **only** the terminal lookup still preserves the response — the outer guard catches it — but loses the log. That is why both layers are here, and the tests tell them apart rather than both asserting "the response is fine".

commons unit 587 · every service's unit suite · all integration · all e2e — green. `kugel_common` bumped to 0.1.77.

## Left on the issue

The issue also notes that `security.get_terminal_info_from_terminal_service` mapping a connection failure onto "Invalid API key" plus an audit entry is arguably its own defect — a dependency being unreachable is not an authentication attempt, and the false entries pollute an audit trail. That is a change to the auth path rather than the logger, and it is not what this PR is about; the logger no longer cares either way.

### DIFF
--- a/services/account/Pipfile
+++ b/services/account/Pipfile
@@ -13,7 +13,7 @@ motor = "*"
 pytz = "*"
 pyjwt = "*"
 passlib = {extras = ["bcrypt"], version = "*"}
-kugel_common = {file = "commons/dist/kugel_common-0.1.76-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.77-py3-none-any.whl"}
 
 bcrypt = "==3.2.0"
 python-multipart = "*"
--- a/services/account/Pipfile.lock
+++ b/services/account/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "8c7a4e6e8d9f206486c8504b0fc41c17ab873aff5b63af9c3a5b165b64a1da76"
+            "sha256": "38092c491d5c8f7783fac8038f0cb433bfcbbaf9b59c6e55c6b663c60ef089a2"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -667,9 +667,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.76-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.77-py3-none-any.whl",
             "hashes": [
-                "sha256:480292c99322b0fe4bcb9eda5332aad2de09c29a58c19dddfc4b36b1e7d1797a"
+                "sha256:7033b63df54f2a7828778048297dea49071ff047c3f56cde362780f8d711de06"
             ]
         },
         "motor": {
--- a/services/cart/Pipfile
+++ b/services/cart/Pipfile
@@ -12,7 +12,7 @@ pydantic-settings = "*"
 motor = "*"
 pytz = "*"
 pyjwt = "*"
-kugel_common = {file = "commons/dist/kugel_common-0.1.76-py3-none-any.whl"}
+kugel_common = {file = "commons/dist/kugel_common-0.1.77-py3-none-any.whl"}
 httpx = "*"
 aiohttp = "*"
 wcwidth = "*"
--- a/services/cart/Pipfile.lock
+++ b/services/cart/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "3059613bb207985a6d07db87340d973da4876e7e6ddf9e952e6ac7a3ecc5ee10"
+            "sha256": "abfab8114fff5519b28b6b161041d479d30e14be16b87b8c2953eb0edcef7a59"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -659,9 +659,9 @@
             "version": "==3.19"
         },
         "kugel-common": {
-            "file": "commons/dist/kugel_common-0.1.76-py3-none-any.whl",
+            "file": "commons/dist/kugel_common-0.1.77-py3-none-any.whl",
             "hashes": [
-                "sha256:480292c99322b0fe4bcb9eda5332aad2de09c29a58c19dddfc4b36b1e7d1797a"
+                "sha256:7033b63df54f2a7828778048297dea49071ff047c3f56cde362780f8d711de06"
             ]
         },
         "motor": {
--- a/services/commons/src/kugel_common/__about__.py
+++ b/services/commons/src/kugel_common/__about__.py
@@ -14,5 +14,5 @@
 # SPDX-FileCopyrightText: 2024-present kugel-masa <masa@kugel.cloud>
 #
 # SPDX-License-Identifier: MIT
-__version__ = "0.1.76"
+__version__ = "0.1.77"
 
--- a/services/commons/src/kugel_common/middleware/log_requests.py
+++ b/services/commons/src/kugel_common/middleware/log_requests.py
@@ -123,10 +123,7 @@ def log_requests(service_name: str = "NO_SERVICE_NAME"):
                 # it raised before the route was ever called, and the caller was
                 # answered with the logger's error (issue #161). The record is
                 # rebuilt from what is known in _record_request.
-                logger.error(
-                    f"Could not capture the request for logging: {request.method} {request.url}: {e}",
-                    exc_info=True,
-                )
+                _report(f"Could not capture the request for logging: {request.method} {request.url}: {e}")
                 request_info = None
             response = await call_next(request)
             process_time_ms = int((time.time() - start_time) * 1000)
@@ -140,13 +137,23 @@ def log_requests(service_name: str = "NO_SERVICE_NAME"):
             try:
                 await _record_request(request, service_name, accept_time, request_info, response, process_time_ms)
             except Exception as e:
-                logger.error(
-                    f"Request log could not be written for {request.method} {request.url}: {e}",
-                    exc_info=True,
-                )
+                _report(f"Request log could not be written for {request.method} {request.url}: {e}")
         return response
 
     return middleware
+
+
+def _report(message: str) -> None:
+    """Report a failure of the logging path, and never become one.
+
+    `logger.error` is not free of risk here: a handler that raises - a full disk,
+    a custom handler - propagates, and these calls sit on paths whose whole
+    purpose is that nothing on them reaches the caller.
+    """
+    try:
+        logger.error(message, exc_info=True)
+    except Exception:
+        pass
 
 
 async def _resolved_or_none(build):
@@ -176,9 +183,12 @@ async def _record_request(
     can reach the caller. Kept as its own function so that boundary is a single
     place rather than a block someone can extend past.
     """
-    is_terminal_service = service_name == "terminal"
-    terminal_info = await _get_terminal_info(request, is_terminal_service)
-    user_dict = await _get_current_user(request)
+    # Both are resolved inside the assembly guard below rather than ahead of it.
+    # Each is internally guarded, but an unexpected failure here used to abort
+    # _record_request outright - no record at all, exactly when one is most
+    # wanted.
+    terminal_info = None
+    user_dict = None
 
     if request_info is None:
         # _make_request_info did not get to return - reading or sanitizing the
@@ -194,6 +204,9 @@ async def _record_request(
         )
 
     try:
+        is_terminal_service = service_name == "terminal"
+        terminal_info = await _get_terminal_info(request, is_terminal_service)
+        user_dict = await _get_current_user(request)
         request_log = RequestLog(
             tenant_id=await _make_tenant_id(terminal_info, user_dict),
             client_info=await _make_client_info(request),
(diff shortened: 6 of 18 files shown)


kugel-masa:
## Review follow-up

Reviewed by Gemini (`gemini-3.6-flash`). It found something I had missed about my own claim, which is the useful kind of finding.

### The second property held only where I had guarded

This PR claims two things: the response is never changed, and **the request is still recorded**. The first is now airtight. The second was not — it held for the two failures I had specifically guarded, and for nothing else.

Every field the record is assembled from can raise on its own, and each of them cost the whole row. The outer guard catches it, so the caller is answered correctly and the audit trail simply has nothing. Two cases it named, both real:

- `_make_response_info` re-reads the response body; a body that will not re-read raised inside the assembly
- `_make_client_info` reached into `request.client` unguarded, and an ASGI scope does not always carry one

Guarded **once at the assembly** rather than six times inside it. If any field cannot be resolved, a minimal record is written instead — the request, the status, and a marker. The point is that the request is recorded, not that every part of it was resolvable, and a field added later is covered without anyone remembering to.

### The failed capture now says so

`body: None` cannot be told apart from a request that legitimately had none, and a reader of an audit trail has only what is written down. It is now `{"_capture_failed": True}`.

### Declined

**`_make_user_info` can raise on missing claims.** The mechanism is real — `RequestLog.UserInfo` rejects `None`, verified — but not reachable: `verify_token` refuses a token without `username` or `tenant_id`, and `is_superuser` defaults to `False`. The assembly fallback covers it regardless, which is a better answer than defaulting each field into a shape that hides a genuinely malformed token.

**A non-JSON request body is recorded as `None`, losing the payload.** True, and deliberate since #155: only JSON bodies have ever been logged. Capturing raw bytes is a decision about that issue's body budget — where the whole point was bounding what a request can write into the log — not about this one.

**Type annotations would fail mypy.** The repository does not run it. The signature this PR introduces is annotated honestly; the rest is left alone rather than half-migrated.

### Verification

|---|---|
| unguard the terminal lookup | 2 |
| remove the guard around the `finally` body | 1 |
| unguard the request capture | 1 |
| drop the `request_info` fallback | 1 |
| remove the assembly fallback | 2 |
| leave the failed capture indistinguishable from no body | 1 |

commons unit 590 · every service's unit suite · all integration · all e2e — green. The live refused-API-key check re-run against the rebuilt stack: `HTTP 401`, logged, recorded status 401.

kugel-masa:
## Second review pass

Reviewed again by Gemini (`gemini-3.6-flash`). Two findings, both real, fixed in `5662f91`. No critical.

### The fallback discarded a tenant it already had

`terminal_info` and `user_dict` are resolved *before* the assembly, so when a later field failed there was no reason to lose them.

Dropping the tenant is not a cosmetic loss: **the buffer routes by it**, so a record without one never reaches the tenant's own database — the audit trail that an auditor actually reads. And a row with no tenant reads as an unauthenticated request, which is a different thing entirely. Same for the client address.

Now both are carried into the fallback, each able to give up on its own without taking the record with it.

### The terminal-lookup warning threw away its traceback

That warning exists to say *why* the terminal could not be resolved. Without `exc_info` it says only that it could not.

### Also: the two markers are now distinct

`_capture_failed` on the request body means the body could not be read. The fallback record now says `_assembly_failed` on the response body, because "the record could not be assembled" is a different question an operator asks — and one string answering both is queryable for neither.

### Answers worth recording

Asked whether the fallback `RequestLog` can itself raise, it walked every field and both paths through `_record_request` and found no input that produces an invalid record. That was the thing I most wanted checked, since a fallback that can fail is worse than none.

Asked what an operator no longer learns, it named the real cost honestly: these failures used to arrive as 500s on client-side HTTP metrics and now appear only in server logs. That is the intended trade — a card charge must not fail because the audit buffer is unreachable — but it does mean the warnings need to be watched.

### Verification

|---|---|
| unguard the terminal lookup | 2 |
| remove the guard around the `finally` body | 1 |
| unguard the request capture | 1 |
| drop the `request_info` fallback | 1 |
| remove the assembly fallback | 2 |
| leave the failed capture indistinguishable from no body | 1 |
| fallback discards the resolved tenant | 1 |
| unguard `request.client` | 1 |
| reuse one marker for both failures | 1 |

The `request.client` guard had been asserted only through a mock of the function containing it — it now has a test that calls it with a scope that has no client.

commons unit 592 · every service's unit suite · all integration · all e2e — green.

kugel-masa:
## Third review pass

Reviewed again by Gemini (`gemini-3.6-flash`). Four findings, all verified before acting, all fixed in `268ba89`.

### `logger.error` could escape

Confirmed with a handler that raises `OSError`:

```
a failing log handler : PROPAGATES (OSError)
```

These calls sit inside guards whose whole purpose is that nothing on them reaches the caller — and one of them runs **before** `call_next`, so it could fail a request outright. A report that becomes the failure it was reporting is the same defect one level down. `_report` cannot.

### Resolving the terminal and the user sat outside the assembly guard

Both are internally guarded, but an unexpected failure aborted `_record_request` outright: **no record at all**, exactly when one is most wanted. Moved inside, so a failure there costs only the fields it resolves.

### The file log could cost the database record

It ran unguarded before the buffer write, so a local disk problem took the copy that is actually queried.

### `service_name` was never stored

The most interesting one. The middleware has passed it since the beginning and `RequestLog` never declared the field, so **Pydantic dropped it silently**:

```
service_name kept on the model : <<ABSENT>>
service_name in model_dump     : False
```

And no stored document carries it — the top-level keys of a real row are `_id, shard_key, created_at, updated_at, etag, tenant_id, client_info, request_info, response_info, staff_info, user_info, terminal_info, snapshot_info`.

Every service writes to the same `{prefix}_commons` collection, so a reader could not tell cart from journal. Declared now; the live check reads back `service_name: cart`.

This one is adjacent to the issue rather than part of it. It is a two-line field on the record this PR spends its whole length making trustworthy, and leaving a known silently-dropped field inside it would be odd — but it is worth naming as scope I added rather than scope I was given.

### Verification

Thirteen mutations across three rounds, each failing a distinct test. This round:

|---|---|
| let the reporter raise | 1 |
| drop the `service_name` field | 2 |
| resolve the context outside the assembly guard | 1 |
| unguard the file log | 1 |

commons unit 597 · every service's unit suite · all integration · all e2e — green. Live: `HTTP 401`, logged, status 401, `service_name: cart`.

### Where three rounds landed

Round 1 found three real defects, round 2 two, round 3 four — the count did not fall, but what they were about did: rounds 1 and 2 were about the response and the record surviving, round 3 was about the guards themselves being trustworthy and the record being readable. Two of this round's four were in code the earlier rounds had already been over.

PR #189, opened 2026-08-22
test(cart): replaying a pre-bill snapshot after finalize records one sale (#152)

### LINKED ISSUE #152: cart: idempotent transaction finalization keyed on cart_id (replay double-count guard)

kugel-masa:
Closes #152.

## What was left

#152's three code items landed with #156/#157 and #172 — `cart_id` on the tranlog, the partial-unique index on `(tenant_id, store_code, cart_id)`, and idempotent finalization. Verified against the running stack before writing anything:

```
first bill        : HTTP 200  transactionNo=7303
replayed bill     : HTTP 200  transactionNo=7303
tranlogs for this cart_id: 1
```

What the issue asked for and did not have was the **test**. #172's suite races two finalizes at once; the scenario #152 is named for is sequential — a snapshot taken while the cart was still `paying`, presented again long after the cart is billed and gone from the cache. Its signature is valid, it is not itself a terminal-state cart, and the server has nothing to compare it against.

## Five tests

- the sale is recorded once — SC-004 of #148
- the replay is answered with the transaction that exists, **not refused**: a terminal replays because it did not hear the first answer, and telling it the request is invalid leaves it holding a sale it cannot account for
- a replay claiming a different seq and counter still records one sale, and gets back what was recorded rather than what it asked for
- a cancel of an already-billed cart **is** refused — it is not a retry of a sale
- the recorded transaction names its cart, which nothing else would notice if it stopped doing

End to end because what is asserted is the row count in the transaction log, and the dedupe lives in an index.

## One expectation of mine was wrong

The third started as an assertion that a replay claiming different numbers is refused. It is not: `__is_same_finalize` compares `(transaction_type, is_cancelled)`, so the identity is the cart and the operation, not the numbers carried with it.

The behaviour is defensible — recording it would be the double-count, refusing it would strand the terminal — so the test now pins it rather than the expectation I brought to it. Worth knowing when reading the log: **nothing on the server tells this apart from an identical retry.** The terminal can see it, because the numbers it gets back are not the ones it sent.

## What the mutation testing showed

Mutation-checked against the **running container**, since these talk to the live service.

Neither dedupe layer alone fails these tests, because there are two: a pre-check before the insert, and a recovery after the unique index rejects it. Each covers the other for the sequential case.

| mutation | result |
|---|---|
| disable the pre-check | 5 passed — the recovery path catches it |
| disable the recovery | 5 passed — the pre-check catches it |
| **disable both** | **3 of 5 fail** |

That is the honest way to state what they cover: the dedupe as a mechanism, not either implementation of it.

All e2e — green.

### DIFF
--- a/services/cart/tests/e2e/test_replayed_finalize.py
+++ b/services/cart/tests/e2e/test_replayed_finalize.py
@@ -52,15 +52,25 @@ async def _tenant_is_set_up(http_client):
 
 
 @pytest_asyncio.fixture(autouse=True)
-async def _remove_synthetic_transactions():
-    yield
+async def carts_created():
+    """Collect this suite's carts, and take only those back out afterwards.
+
+    Deleting by a `transaction_no >= base` floor is how the neighbouring suites
+    do it, and it reaches across them: the bases are 9000, 9500 and 9600, so the
+    lowest floor deletes everything above it. Harmless while the suites run one
+    after another, and not something to leave for whoever runs them in parallel.
+    """
+    created = []
+    yield created
+
     from kugel_common.database import database as db_helper
 
     db = await db_helper.get_db_async(f"{os.environ.get('DB_NAME_PREFIX')}_{os.environ.get('TENANT_ID')}")
-    await db["log_tran"].delete_many({"transaction_no": {"$gte": SYNTHETIC_SEQ_BASE}})
+    if created:
+        await db["log_tran"].delete_many({"cart_id": {"$in": created}})
 
 
-async def _cart_ready_to_bill(http_client, terminal_id, header):
+async def _cart_ready_to_bill(http_client, terminal_id, header, created=None):
     """Take a cart to `paying` and return its cart_id with the snapshot from there."""
     response = await http_client.post(
         f"/api/v1/carts?terminal_id={terminal_id}",
@@ -69,6 +79,8 @@ async def _cart_ready_to_bill(http_client, terminal_id, header):
     )
     assert response.status_code == status.HTTP_201_CREATED, response.text
     cart_id = response.json()["data"]["cartId"]
+    if created is not None:
+        created.append(cart_id)
 
     response = await http_client.post(
         f"/api/v1/carts/{cart_id}/lineItems?terminal_id={terminal_id}",
@@ -116,9 +128,9 @@ def _finalize(snapshot, seq, counter):
 
 
 @pytest.mark.asyncio
-async def test_the_sale_is_recorded_once(http_client, api_header, opened_terminal_id):
+async def test_the_sale_is_recorded_once(http_client, api_header, opened_terminal_id, carts_created):
     """SC-004 of #148: exactly one recorded transaction per cart_id."""
-    cart_id, paying = await _cart_ready_to_bill(http_client, opened_terminal_id, api_header)
+    cart_id, paying = await _cart_ready_to_bill(http_client, opened_terminal_id, api_header, carts_created)
     url = f"/api/v1/carts/{cart_id}/bill?terminal_id={opened_terminal_id}"
     request = _finalize(paying, SYNTHETIC_SEQ_BASE + 1, 61)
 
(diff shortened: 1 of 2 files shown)


kugel-masa:
## Review

Reviewed by Gemini (`gemini-3.6-flash`). Three findings, all on tests that had just been written to check something else — a fair place to be caught. Fixed in `50732a7`.

### A refusal was asserted as "not 200"

`test_a_different_operation_on_the_same_cart_is_refused` checked only `status_code != 200`. With both dedupe layers gone the unique index refuses the insert too, and **the 500 from that satisfies `!= 200` while saying the opposite about the code under test**. Now the deliberate `409 CONFLICT` and its error code `401511`.

### A body was read before its status

`test_the_replay_is_answered_with_the_transaction_that_exists` went straight to `.json()["data"]`. If the replay were refused rather than answered idempotently, it failed with a `KeyError` on `data` instead of saying what happened.

### The teardown deleted by a floor other suites share

The bases in this directory are 9000, 9500 and 9600, and each teardown deletes everything at or above its own — so the lowest reaches across all of them. Harmless while the suites run one after another (the runner is plain `pytest -m e2e`, no xdist) but not something to leave for whoever runs them in parallel. This suite now collects the carts it created and takes back only those.

### Also added: why the dedupe index is partial

The review noticed the docstring claims something no test checks. Transactions predating `cart_id` have none, and the index is unique — so without the partial filter every historical row would collide with every other on a missing key, and the collection could never be indexed at all. Nothing would notice until a tenant with history could not be set up, which is what #185 looks like from the outside.

Two integration tests: rows without a `cart_id` coexist, and two rows sharing one still collide. Mutation-checked by dropping the partial filter.

### Not changed

The finding that neither dedupe layer alone is pinned stands, and is stated in the PR body rather than fixed. Two layers covering each other is the design; a test that fails when only one is removed would be asserting the implementation rather than the behaviour.

All integration · all e2e — green.

PR #191, opened 2026-08-22
feat(cart): record when a repeated finalize carries different numbers (#190)

### LINKED ISSUE #190: cart: a replayed finalize carrying different numbers is not told apart from an identical retry
Found while writing the #152 tests (PR #189).

## The behaviour

`cart_id` is the transaction identity: a second finalize for the same cart returns the transaction already recorded instead of writing another. That is right — a terminal replays because it did not hear the first answer, and refusing it would leave a sale it cannot account for.

But "the same finalize" is decided on two fields only:

```python
# TranlogRepository.__is_same_finalize
existing.transaction_type == incoming.transaction_type
and self.__is_cancelled(existing) == self.__is_cancelled(incoming)
```

The numbers the terminal carried are not compared, and not recorded. So these two are indistinguishable afterwards:

| | first attempt | replay |
|---|---|---|
| **A. an honest retry** | seq=100, receiptCounter=41 | seq=100, receiptCounter=41 |
| **B. the terminal advanced** | seq=100, receiptCounter=41 | seq=101, receiptCounter=42 |

Both produce one tranlog holding `seq=100, receiptCounter=41`, and one log line:

```
Idempotent finalize: tranlog for cart_id=X already exists
```

The values the replay actually carried (101 / 42) are written nowhere.

## Why B matters

The terminal's running receipt counter has moved past what the server recorded, which means one of:

- **it printed receipt 42** — a number that appears in no transaction log, on a receipt a customer is holding
- **the counter advanced without a print** — a gap, which #166 explicitly allows (gaps expected, reuse forbidden)
- **a different transaction reused the cart_id** — a client defect

The first is worth investigating and the second is not, and nothing recorded distinguishes them. `transactionDatetime` can diverge too; usually by seconds, but across midnight it changes the business-date bucket the transaction is counted in.

The terminal itself can see the divergence, because the numbers it gets back are not the ones it sent. What it does about that is client-side, and leaves no trace here.

## What to record

When the pre-check returns an existing tranlog, compare the carried finalize context against the recorded one and, when they differ, say so:

```
Idempotent finalize with a divergent context: cart_id=X
  recorded: seq=100 receipt_counter=41 receipt_no=111141 generate_date_time=...10:30:00
  carried : seq=101 receipt_counter=42 receipt_no=111142 generate_date_time=...10:30:07
```

Plus an audit record, following what #168 already does for the other numbering divergence: an ERROR line and a row through `__add_restore_audit_async` with a `result` naming this case. That gives a query for "replays whose receipt counter had moved on", which is the set worth reconciling against printed receipts.

## Scope

Detection only. The current handling is correct and should not change: the sale stays recorded once, and the caller still gets the numbers that were actually recorded.

## References

- `services/cart/app/models/repositories/tranlog_repository.py` — `__is_same_finalize`, and the pre-check in `create_tranlog_async`
- `services/cart/app/services/cart_service.py` — `__audit_numbering_fallback_async`, the #168 precedent
- `services/cart/app/api/v1/schemas.py` — `FinalizeContext` (seq / receipt_no / receipt_counter / transaction_datetime)
- `services/cart/tests/e2e/test_replayed_finalize.py::test_a_replay_claiming_other_numbers_still_records_one_sale` — pins the current behaviour
- #152 / #166 / #168

kugel-masa:
Closes #190.

## What it changes

A second finalize for the same `cart_id` is answered with the transaction already recorded rather than writing another — right, and pinned by #152. What the log did not say is whether the repeat carried the **same numbers**.

| | first attempt | repeat | recorded |
|---|---|---|---|
| **an honest retry** | seq=100, counter=41 | seq=100, counter=41 | 100 / 41 |
| **the terminal advanced** | seq=100, counter=41 | seq=101, counter=42 | 100 / 41 |

Both produced one tranlog and one identical log line. The values the repeat carried were written nowhere.

Now the two are told apart. An identical repeat stays a warning; a divergent one is an ERROR naming every field that differs with both values, plus an audit row through the same trail #168 writes its numbering divergence to — so the set worth reconciling against printed receipts is a query rather than a reading of logs.

Live:

```
identical repeat             HTTP 200  audit rows: 0
repeat with other numbers    HTTP 200  audit rows: 1
    reason: transaction_no: carried 8527 vs recorded 8526,
            receipt_no: carried 111112 vs recorded 111111,
            receipt_counter: carried 2 vs recorded 1
    diverged flag: True
```

## Which numbers

The four the terminal carries in its finalize context (#156): `transaction_no` (from `seq`), `receipt_no`, `receipt_counter`, `generate_date_time`.

`receipt_counter` is the one that matters most — it is what a printed receipt number is derived from, so a terminal ahead of the record may have handed a customer a number that appears in no transaction log. `generate_date_time` usually differs by seconds and matters when it crosses midnight, because that is the business date the transaction is counted in.

## Detection only

As the issue asks. What the caller gets back does not change: the sale stays recorded once, and the response still carries the numbers that were actually recorded. An audit write that fails is swallowed — the transaction is recorded and answered, and losing the note about it must not turn that into a failure.

Both repeat paths report: the idempotent pre-check and the concurrent-race recovery, so an ordinary retry and the #172 race are told apart the same way.

## Verification

|---|---|
| never see a divergence | 1 |
| audit every repeat, divergent or not | 1 |
| compare only the transaction number | 1 |
| let an audit failure escape | 1 |

cart unit 648 · cart integration 76 · all e2e — green.

## Related

This is the last open item under the #146 umbrella that is code rather than a prerequisite: its design table settles replay as *"accept + detect: idempotent finalization keyed on cart_id + audit trail"*, and this is the half of the audit trail that was missing. What remains there is demoting the server-side cache (blocked on clients migrating to `CART_REQUEST_SNAPSHOT_MODE=REQUIRED`) and measuring real cart sizes.

### DIFF
--- a/services/cart/app/dependencies/get_cart_service.py
+++ b/services/cart/app/dependencies/get_cart_service.py
@@ -169,7 +169,8 @@ async def __get_cart_service_async(
     tranlog_repo = TranlogRepository(db=db, terminal_info=terminal_info)
     await tranlog_repo.initialize()
     tranlog_delivery_status_repo = TranlogDeliveryStatusRepository(
-        db=db_common, terminal_info=terminal_info  # use common db
+        db=db_common,
+        terminal_info=terminal_info,  # use common db
     )
     await tranlog_delivery_status_repo.initialize()
     transaction_status_repo = TransactionStatusRepository(db=db, terminal_info=terminal_info)
@@ -203,6 +204,7 @@ async def __get_cart_service_async(
         payment_master_repo=payment_master_repo,
         transaction_status_repo=transaction_status_repo,
         store_info_repo=store_info_repo,
+        cart_restore_log_repo=cart_restore_log_repo,
     )
 
     return CartService(
--- a/services/cart/app/services/tran_service.py
+++ b/services/cart/app/services/tran_service.py
@@ -191,6 +191,10 @@ class TranService:
         # downstream cart_id dedupe converges to one record. Without a carried
         # time (legacy / no-snapshot path) the server-side numbering is used.
         carried = cart.transaction_datetime is not None
+        # A repeat is reported once. The insert path can report and then fail to
+        # commit, which lands in the recovery below - and reporting the same
+        # repeat twice doubles every row the audit query is meant to count.
+        reported = False
         if carried:
             # (business_counter, seq) composite: transaction_no carries seq.
             tranlog.transaction_no = cart.seq
@@ -324,12 +328,13 @@ class TranService:
                 payload=event_message,
                 services=event_distinations,
             )
-            carried = tranlog
+            submitted = tranlog
             tranlog = await self.tranlog_repository.create_tranlog_async(tranlog)
-            if tranlog is not carried:
+            if tranlog is not submitted:
                 # Not a fresh insert: the repository returned a tranlog that was
                 # already there, so this finalize was a repeat (issue #152).
-                await self.__report_finalize_repeat_async(carried, tranlog)
+                await self.__report_finalize_repeat_async(submitted, tranlog, carried)
+                reported = True
             await self.tranlog_repository.commit_transaction()
 
         except FinalizeConflictException:
@@ -350,7 +355,9 @@ class TranService:
             # of a 500. The winner already published, so we do NOT publish again.
             existing = await self.__recover_concurrent_finalize(tranlog)
             if existing is not None:
-                await self.__report_finalize_repeat_async(tranlog, existing)
+                if not reported:
+                    await self.__report_finalize_repeat_async(tranlog, existing, carried)
+                    reported = True
                 return self.__apply_tranlog_to_cart(cart, existing)
             # The duplicate was on some other unique index (not the cart_id race) —
             # surface it as a real failure.
@@ -372,7 +379,9 @@ class TranService:
                     tranlog.cart_id,
                     type(e).__name__,
                 )
-                await self.__report_finalize_repeat_async(tranlog, existing)
+                if not reported:
+                    await self.__report_finalize_repeat_async(tranlog, existing, carried)
+                    reported = True
                 return self.__apply_tranlog_to_cart(cart, existing)
             message = f"Error creating tranlog: {e}"
             raise InternalErrorException(message, logger) from e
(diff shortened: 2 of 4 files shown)


kugel-masa:
## Review follow-up

Reviewed by Gemini (`gemini-3.6-flash`). Four findings, all acted on in `ce7cbc5`. One of them was a false-positive I had shipped, and another was that my tests could not have caught it.

### The server's own numbering was filed as a terminal divergence

On the server-numbered path the numbers in hand were issued by **this** request from the server's counter and clock, so two simultaneous bills differ by construction. Measured on the running stack before the guard:

```
two concurrent server-numbered bills: 200 / 200
tranlogs: 1
FALSE divergence rows: 1
  transaction_no: carried 50 vs recorded 49,
  receipt_no: carried 111160 vs recorded 111159,
  generate_date_time: ...419981 vs ...419723
```

Nothing there is a terminal that moved on. Filing it buries the rows that mean something — the opposite of what this issue is for.

The check now runs only when the client carried the finalize context. The method already had that flag (`carried = cart.transaction_datetime is not None`) — and my new local shadowed it, which is a fair part of why I did not see this.

### The tests would not have caught it

They called the reporter directly through name mangling, so **removing every call to it left them all passing**. Three e2e tests now go through a real finalize: a divergent repeat leaves a row, an identical one does not, and a server-numbered race does not.

Mutation-checked against the running container:

| mutation | result |
|---|---|
| never report a repeat (remove the call) | `test_a_repeat_carrying_other_numbers_is_written_down` fails |
| report on the server-numbered path too | `test_a_server_numbered_race_is_not_written_down` fails |

### A repeat could be reported twice

The insert path reports and then commits; a commit that fails lands in the recovery, which reports again. Two rows for one repeat doubles exactly what the audit query is meant to count.

### The same instant written two ways read as a divergence

The carried time is validated as ISO-8601 and not normalised, so `+00:00` against `Z`, or trailing zero microseconds, compared unequal. Parsed before comparing.

### Not changed

**`tranlog is not carried` relies on reference identity across the repository boundary.** Fair as a fragility, but the alternative — returning `(tranlog, created)` — changes an API two call sites use, and a repository that started cloning would break more than this. The unit suite pins that a fresh insert reports nothing, so a change of that kind fails loudly rather than silently filing every checkout as a repeat.

cart unit 653 · cart integration 76 · all e2e — green.

kugel-masa:
## Second review pass

Reviewed again by Gemini (`gemini-3.6-flash`). **One finding**, minor, and no critical or major — it walked the `reported` flag through every path in the finalize block, the `client_numbered` proxy against cancels and partial contexts, `__same_instant` against naive-vs-aware datetimes, and the audit write's placement relative to the MongoDB transaction, and found nothing in any of them.

### The concurrent-race recovery was not covered

Correct. It is a separate call site from the idempotent pre-check, and the e2e reaches only the sequential repeat, where the pre-check is what answers. Four unit tests now drive `create_tranlog_async` itself through both race shapes — the unique index rejecting at the insert, and at the commit. Mutation-checked: removing the recovery reports fails both.

### A correction to my own previous comment

I wrote that reporting a repeat twice "doubles exactly what the audit query is meant to count". **It does not**, and writing the test for it is what showed that:

```
[ERROR]   Finalize repeated for cart_id=cart-190 carrying DIFFERENT numbers: {...}
[WARNING] Finalize failed but the same finalize is already persisted; returning it
[WARNING] Finalize repeated for cart_id=cart-190 carrying the same numbers
```

By the time the recovery reports, the tranlog in hand **is** the recorded one — so the second comparison is the record against itself, finds no divergence, and no second audit row is possible on that path.

What the second report does produce is that third line: the same repeat described as carrying the same numbers, directly after a line saying it carried different ones. The contradiction is in the log rather than the audit trail, and a reader has only what is written down.

The guard stays. The reason for it is narrower than I claimed, and the test now asserts the **line count** rather than the audit count — which is what the mutation actually moves. It did not bite before precisely because it was asserting the thing that was never at risk.

### Verification

|---|---|
| never see a divergence | 1 |
| audit every repeat, divergent or not | 1 |
| compare only the transaction number | 1 |
| let an audit failure escape | 1 |
| never report a repeat (remove the call) | 1 e2e |
| report on the server-numbered path too | 1 e2e |
| the recovery paths stop reporting | 2 |
| drop the once-only guard | 1 |

cart unit 657 · cart integration 76 · all e2e — green.

PR #196, opened 2026-08-22
feat(cart): one cart, one path — a carried cart is never written to the cache (#192)

### LINKED ISSUE #192: cart: mixing the carried and cache paths on one cart silently loses its contents; the creation-time cache write is what makes it possible
From a comprehensive external review of the client-carried cart design (#146). Part of that umbrella.

## What happens

In DUAL mode the path is chosen **per request**, on one condition only — whether a snapshot came with it (`get_cart_service.py:83`):

```python
snapshot = request.scope.get("cart_snapshot")
if snapshot is not None:
    await cart_service.prepare_stateless_from_snapshot(...)
```

The carried path deliberately writes nothing (`cart_service.py:1077`):

```python
if self._stateless:
    # the response snapshot is the authority; do not depend on server-side cache (FR-004)
    self._snapshot_cart = cart_doc
    self.current_cart = None
    return          # cart_repo is never reached
```

So the cache holds the cart as of the last cache-path write — which, for a client that carries, is **cart creation**. One snapshot-less request then continues from there:

```
create                 → cache: empty cart
add item A  (carried)  → cache unchanged
add item B  (carried)  → cache unchanged        client holds: A, B
one request without a snapshot
                       → reads the cache: no A, no B
                       → answers with a correctly signed snapshot of a cart missing both
```

No error is raised. The response is a properly signed, structurally complete envelope. Nothing in it tells the terminal that two lines were dropped.

`delete_cart_async` is skipped the same way (`cart_service.py:1173`), so a cart finalized on the carried path also stays in Redis until its TTL — visible to any later snapshot-less request.

## What the design says, and why that is not enough

`docs/en/cart/client-carried-cart.md` records this as accepted:

> switching paths inside one cart rebuilds it from a cache that the stateless path deliberately stopped writing, so the cart itself has rolled back… Every cart route is a body-carrying mutation, so a client that carries carries throughout.

That last sentence is an **expectation of the client**, not something the server enforces — and the detection added in #165 only fires if the client carries *again* afterwards, because the rollback query looks for a revision that goes **down**. A client that stays on the cache path leaves no trace of what was lost.

The cases are ordinary migration-period ones: an endpoint where a client forgot to attach the snapshot, a half-migrated fleet, an error path that retries without one, or two backends on different modes. DUAL exists precisely for that period.

## The root: the creation-time write

`POST /carts` always writes the cart to the cache (`cart_service.py:268`, `isNew=True`) — Redis via the Dapr `cartstore`, MongoDB `cache_cart` on failure. It does so because:

1. creation has no prior snapshot to carry, so the DI cannot arm the stateless path
2. the response then **reads the cart back** to build the envelope (`api/v1/cart.py:93`), described in its own comment as *"built from the cached (authoritative) cart"*

Neither is a requirement of the stateless design. Both exist to serve a client that might **not** carry. In REQUIRED mode the write is pure waste — and it is the source of the stale copy every scenario above depends on.

## Proposed direction

**Have the client declare at creation whether it will carry, and write nothing when it will.**

```
POST /carts   carrying → build the cart in memory, sign it, store nothing
                       → a later snapshot-less request finds no cart → 404
```

The rejection then comes for free: there is nothing to read, so there is nothing stale to read. No flag, no new state, no new branch — **one fewer write** than today, in the direction #146 set out in.

It also removes the residue of finalized carried carts, and the revision false-positive the docs describe, since mixing stops being reachable at all.

Requires: the creation response builds its envelope from the in-memory document rather than reading it back — the same thing the carried path already does with `self._snapshot_cart`.

### The trade-off, stated plainly

The client gets `404 Cart not found`, not "this cart is carried and cannot continue without its snapshot". That reads like the very failure #146 opened with (a timeout turning an in-flight cart into a 404), though it means the opposite here.

Keeping a distinguishable error means remembering that the cart was carried, which means storing something — and storing state to produce a better error message is the wrong way round for an issue about not storing state. The message and the docs can carry that meaning instead.

### Alternatives considered

- **A flag on the cart document, written at creation.** Costs no extra write (creation writes anyway) and gives a precise 409. Rejected: it keeps the stale copy, so it guards the symptom while leaving the cause.
- **Per-terminal mode, from the token or the terminal record.** No cart state at all and no client change, but it moves the truth to terminal registration, where the same mistake produces the same accident.
- **Detect from the request log.** #165 records `snapshot_info.cart_id` with an index for it, so "has this cart ever presented a snapshot" is one query. Not usable as a guard — the log is buffered and up to 5s late — but it is a usable **detector** for finding mixed carts after the fact.

## References

- `services/cart/app/dependencies/get_cart_service.py:83` — the per-request choice
- `services/cart/app/services/cart_service.py:268` — the creation write; `:1077` and `:1173` — the carried path skipping the write and the delete
- `services/cart/app/api/v1/cart.py:93` — the creation response reading the cart back
- `docs/en/cart/client-carried-cart.md` — where this is currently recorded as accepted
- #146 (umbrella), #156 (phase 2), #165 (rollback detection, whose false-positive note this would retire)

kugel-masa:
Closes #192.

## The defect

The path was chosen per request, on whether a snapshot came with it, and nothing tied a cart to one of them. The carried path writes nothing, so a cart built up by carried requests left the cache holding it **as it was at creation** — and one snapshot-less request continued from there:

```
create                 → cache: empty cart
add item A  (carried)  → cache unchanged
add item B  (carried)  → cache unchanged        client holds: A, B
one request without a snapshot
                       → reads the cache: no A, no B
                       → answers with a correctly signed snapshot of a cart missing both
```

No error. Nothing in the log unless the client happened to carry again afterwards, since the rollback query looks for a revision that goes *down*.

The docs recorded this as accepted, on the grounds that *"a client that carries carries throughout"* — an expectation of the client, not something the server enforced.

## The fix, and why it removes a write rather than adding a guard

**The client says which way it will work when it opens the cart**, and nothing is written to the cache when it will carry — not even at creation.

Creation is the one request that always wrote, because it has nothing to carry yet and the server could not know what the client would do. That write is a DUAL-mode artifact rather than a requirement of the design, and it is the source of every stale copy above. Removing it makes the mixture **impossible rather than detectable**: there is no cart to continue from.

Measured on the running stack:

```
carrySnapshot=true    carried → 200      without a snapshot → 404 (401002)
carrySnapshot=false   carried → 409      without a snapshot → 200
```

**Both directions are refused**, because both lose the same thing — declaring the cache path and then carrying leaves the cache copy behind while the cart moves on, which is the same silent loss reached from the other side. The live check before adding that half showed it plainly: a cart carried to two lines, then continued without a snapshot, came back with one.

That refusal needed somewhere to record how the cart was opened, and the answer costs no cache read: **the declaration rides in the signed cart document**, so a snapshot states which way its own cart was opened. A cart created before the field existed says neither and is left alone, so carts in flight across the deployment keep working.

`carrySnapshot` defaults to false — what a client that predates it means by not sending it — so nothing changes for one that does not carry. In `CART_REQUEST_SNAPSHOT_MODE=REQUIRED` no declaration is needed: every mutating request has to carry, so a cached copy could never be read and none is written.

Its own error code (`401515`, 409) rather than borrowing the snapshot-scope one, whose message says the tenant or store did not match.

## Test fixtures

A number of suites created a cart and then carried it — precisely the mixture now refused. They declare the carried path and carry from the first request.

The two that are *about* the server-numbered path keep a cache-path cart, and now say so: `test_legacy_cancel_is_unaffected` and `test_a_server_numbered_race_is_not_written_down`. Both were silently relying on the mixture to set themselves up.

## Verification

|---|---|
| write the cart despite the declaration | `test_a_request_without_its_snapshot_is_refused` (e2e) |
| drop the reverse guard | `test_carrying_it_is_refused` (e2e) |

Both against the running container.

cart unit 666 · every service's unit suite · all integration · all e2e — green.

## What this retires

The "one limit worth stating" paragraph in `client-carried-cart.md`, and with it the revision false-positive it described. Mixing is no longer reachable, so there is no case for the rollback query to explain away.


kugel-masa:
## External review follow-up (codex)

Two P1 findings, both verified against the code. One is fixed here; the other is dismissed on a factual ground.

### Fixed — a carried cart is lost when signing is degraded

`carrySnapshot: true` means the cart is written nowhere. When the signer is unavailable the creation response carries `signedSnapshot: null`, so the client can neither carry the cart nor find it: the next request is a 404. Before the declaration existed the cart was always cached, so a degraded key cost a feature rather than the cart.

Falling back to the cache is not an answer here — the cache path is being retired. The situation is removed instead:

- **The service refuses to start without a usable signing key.** Unset, malformed, and shorter than 32 bytes each stop the process. The key committed to this repository is refused the same way, unless `SNAPSHOT_ALLOW_INSECURE_KEY=true` says the stack is a development one. Reviewers were right that an ERROR log is not enough: it signs and verifies, so every other signal looks healthy while the signature protects nothing.
- **A carried response is never returned unsigned.** Startup cannot see a key that loads and then fails to sign; that case now fails with 503 (`401507`) instead. Nothing is lost by refusing — a carried request writes no cart state, so the client repeats it with the snapshot it still holds, and a finalize is idempotent by `cart_id` (#170).

A cart the server holds in its cache is unaffected: its snapshot stays best-effort and the field goes out null.

While wiring this up: **`docker-compose.prod.yaml` passed no signing key at all**, so the production stack has been running degraded since the feature shipped — no snapshot has ever been issued there. It now requires one explicitly.

### Not changed — omitting `carrySnapshot` still means the cache path

The reviewer's point is correct in principle: a client that already carries snapshots does not send the new field, so its carts record `false` and its first carried request is refused with 409. The `None` compatibility case only covers carts in flight across the deployment.

It does not apply, because no client is carrying snapshots in production yet. Treating omission as "either path is fine" would leave the exact defect #192 exists to close open for the clients least able to detect it, which is the worse trade once there is nothing to protect.

One alternative was considered and rejected as unworkable: pinning the path lazily, on the first request. An undeclared cart is cached at creation, so pinning it to the carried path would have to be written back to that cached copy — restoring the cache write this issue exists to remove. Without the write-back a later snapshot-less request still reads the stale copy, so the defect stays open.

### Verification

- cart: 690 unit, 78 integration, 85 e2e
- all services: unit 8/8, integration 7/7, e2e 8/8
- Both new guards mutation-tested (removed → the tests fail).
- Live: with `SNAPSHOT_ALLOW_INSECURE_KEY=false` the cart container exits at startup right after the key check; with it true the same container runs.

(stock e2e failed once in a batch run on a websocket auth timing issue and passed on its own and on the re-run — unrelated to this change.)

Issue #211, opened 2026-09-03
commons: a request body's secrets are recorded verbatim, in the request log and in app.log


kugel-masa:
Split out of masking work carried in a downstream fork. Reporting it as a gap rather than a patch, because the request logger has no notion of a secret at all — and the one value it does mask (`X-API-Key`, via `mask_dict_api_key`) shows the intent was there.

## What is recorded

`sanitize_log_body` removes fields by name for **size** — `DEFAULT_LOG_STRIP_FIELDS` is `("signedSnapshot", "signed_snapshot")` — and nothing else is filtered. The body reaches two sinks as it arrived:

- the `RequestLog` document (`log_requests.py:559`, `:590`)
- `app.log` at DEBUG, through no filter at all (`log_requests.py:495`):

```python
logger.debug(f"request body: {json_body}")
```

There is no path exclusion anywhere in the middleware.

## A concrete case

`POST /register` and `POST /register/user` take `UserAccount`, whose `password` is plain text (`account/app/api/common/schemas.py:44`). The value is hashed before storage — `password="*****"` at `account.py:150` — but the *request log* keeps what arrived. So the plaintext password sits in the audit collection, and at DEBUG in `app.log`.

`/token` happens to escape: it takes form data, `json.loads` fails, and the body becomes `None`. That is luck rather than a boundary.

## Two properties make this wider than one endpoint

**The body is recorded ahead of validation.** `_make_request_info` runs in the middleware, before FastAPI parses anything, so a request answered with 422 is still logged in full — including a field that never passed a schema.

**Names alone are not enough.** A field whose schema accepts arbitrary string keys — a provider-agnostic credential bag, say — can carry `cardN0` or `pinCod`, and any name-based filter misses them. Masking every value beneath a container field, whatever its key, is what closes that.

## What a fix looks like

A downstream fork carries one. It sits beside `sanitize_log_body` rather than replacing it, since the two answer different questions — size versus secrecy:

- a set of secret field names matched case- and separator-insensitively (`pin`, `password`, `token`, `secret`, `cardNo`, `pan`, `authorization`, `X-API-Key`, `dapr-api-token`, and the rest)
- container fields whose every descendant value is masked regardless of key name
- `X-API-Key` kept partial (`first4…last4`), matching what `mask_dict_api_key` already established
- the same masking applied to validation-error details, which otherwise echo the rejected value straight back

None of it is specific to that fork's domain.

[...]

PR #219, opened 2026-09-04
fix: clear the logging leftovers before the downstream port

kugel-masa:
Three loose ends around credential logging, cleared upstream so the ST-POS port carries a finished state rather than reproducing these.

## 1. A debug line one character from a leak (journal / report / terminal)

```python
logger.debug("OpenCloseLogRepository.create_open_close_log: open_close_log->{open_close_log}")
```

No `f`, so today it writes the braces out literally and says nothing. Add the prefix — the obvious "fix" for anyone who notices it printing `{open_close_log}` — and it dumps an `OpenCloseLog`, which carries `terminal_info` and therefore an api_key and a staff pin. Exactly the shape #211 was about, and the masked line a few statements below already logs the same document properly.

Deleted rather than given the prefix and the masking: it duplicates the line below it, and the safest version of a line that says nothing is no line. The downstream fork has the **f-string** version of this in terminal, unmasked, so the port now has somewhere to land.

## 2. The unused exception-handler twin

`kugel_common/utils/api_exception_handler.py` was a second implementation of the handlers every service registers from `exceptions/exception_handlers.py`, imported by nothing but its own tests. It carried none of the masking — no `mask_validation_error_details` — so a 422 built by it would echo the rejected `pin` straight back. It does not echo the input today only because it formats less than its twin: one edit away, by someone making the two match, from being the defect #211 fixed.

A copy nobody runs is a copy nobody notices going wrong. Deleted, along with its own test file; `test_error_responses_do_not_echo_credentials.py` now covers the one module that exists and records why there is only one. Confirmed unreferenced in both this repo and the fork.

## 3. A hashed credential is credential material

Name matching is exact after normalization, so `secret` in `_SECRET_FIELD_NAMES` does not cover `secret_hash` — a real field on the fork's `EdgeTerminal`, holding bcrypt output. A hash is not the plaintext, but it is offline-crackable and belongs in no log; the point #214 made when it kept `hashed_password` out of the register log.

Named: `secrethash`, `hashedsecret`, `passwordhash`, `pinhash`, `hashedpin`, `apikeyhash`, `hashedapikey`. A hashed api key is blanked outright rather than kept partial — the `first4...last4` form exists for a troubleshooting workflow that has no equivalent for a hash.

Commons **0.1.90**; all seven Pipfiles and locks moved with it (the `kugel_common` entry alone, so nothing unrelated is re-resolved — verified: no lock line outside `kugel_common` / `_meta` / the wheel hash changed).

## Tests

Unit tiers, all green: commons 689, account 45, terminal 373, master-data 406, cart 697, report 229, journal 156, stock 225. Two new masking tests cover the hashed spellings and the api-key-vs-its-hash distinction. Every service's `app.main` imports cleanly against 0.1.90 with the deleted module gone.

Integration and e2e were **not** run: they need a running stack, and the local one was stopped at the end of the previous task. The change is a deleted unused module, a frozenset addition and three deleted log lines, all covered above — but say the word and I will bring a stack up and run them.

External review (codex): no required changes; it checked the reference removal, the lock consistency across the seven services, and that the added hash names do not collide with the `apiKey` partial-mask rule.

## Docs

Rule 1 of the masking section (ja/en) now names the hashed spellings and says why exact matching makes that necessary.

https://claude.ai/code/session_01354h8PdbZCuE1BQkS5eLxd


kugel-masa:
## Integration and e2e, on a stack rebuilt from this branch

All seven images rebuilt (`./scripts/build.sh`) and the stack restarted, so nothing is left over from an earlier build.

**Integration — 7/7 services, 0 failures.**

**E2E — 8/8 suites, 245 tests, 0 failures:**

| suite | passed |
|---|---|
| account | 4 |
| master-data | 19 |
| terminal | 9 |
| journal | 6 |
| stock | 15 |
| report | 31 |
| cart | 86 |
| cross-service | 75 |

The cross-service suite includes `test_credential_sentinels.py`, which ran and passed — it plants per-run sentinels, walks the system including the failure paths, then scans every container's logs and every Mongo collection for them. That is the empirical check that the three deleted log lines and the masking additions did not leave a credential readable anywhere, and it is the one that would have caught the `OpenCloseLog` dump had the `f` prefix been there.

Two notes on the environment, for the record:

- The local stack was previously a mixture of two codebases (the downstream fork's images share these tag names), which is why an earlier run of this suite could not deliver the open/close log to journal and report — the fork subscribes to `topic-*-cloud` on a different route. The rebuild above made it consistent, and the fan-out the sentinel scan waits for now completes.
- The five `test_terminal_counter_rollover.py` integration failures seen earlier were a leftover `db_cart_test_tenant` database colliding case-insensitively with this file's hard-coded `TEST_TENANT` (MongoDB refuses `DatabaseDifferCase`). The downstream fork's copy of the same file uses the lowercase spelling against the same MongoDB, so running its suite recreates the collision. Dropping the stale database cleared it; worth renaming one side to stop it recurring.

Issue #17, opened 2025-10-21
perf: optimize RequestLog middleware to prevent API latency spikes


kugel-masa:
## Problem

The cart AddItem API experiences performance spikes with ~1% of requests taking over 1 second. Analysis indicates that the RequestLog middleware's synchronous database writes are blocking HTTP responses.

## Root Cause Analysis

**Location**: `services/commons/src/kugel_common/middleware/log_requests.py:86-87`

The middleware currently executes database writes **synchronously in the critical path**:

```python
finally:
    await _output_request_log_to_file(request_log)
    await _output_request_log_to_db(request_log)  # ← BLOCKS response
return response
```

### Performance Bottlenecks

1. **Sequential Database Writes** (lines 139-146)
   - Writes to 2 databases sequentially:
     - `{DB_NAME_PREFIX}_commons`
     - `{DB_NAME_PREFIX}_{tenant_id}`
   - Each write involves DB connection, repository instantiation, and insert operation

2. **Blocking I/O on Critical Path**
   - The `finally` block executes **before** returning HTTP response
   - Client must wait for both DB writes to complete
   - Under load, MongoDB write latency increases → API latency spikes

3. **No Error Isolation**
   - If MongoDB is slow/unavailable, **all API requests slow down**

### Why 1% Spike Over 1 Second?

- MongoDB write conflicts under concurrent load
- Connection pool exhaustion during peak traffic
- Replica set lag causing slower writes
- Index building or background operations in MongoDB

## Proposed Solution: Fire-and-Forget Logging

Implement asynchronous logging using `asyncio.create_task()`:

```python
finally:
    # Log to file synchronously (fast operation)
    await _output_request_log_to_file(request_log)
    
    # Log to database asynchronously (fire-and-forget)
    asyncio.create_task(_output_request_log_to_db_async(request_log))
return response  # Return immediately without waiting for DB
```

### Benefits

- **Immediate response**: API returns without waiting for DB writes
- **Reduced P99 latency**: Expected drop from >1s to <100ms
- **Better error isolation**: DB issues don't affect API response time
- **Low risk**: Non-breaking change, maintains same functionality

### Implementation Details

1. Rename `_output_request_log_to_db` to internal function
2. Create wrapper `_output_request_log_to_db_async` with exception handling
3. Use `asyncio.create_task()` to run in background
4. Ensure proper error logging for failed writes

## Testing Plan

1. Performance test with locust (existing test suite)
2. Compare P99 latency before/after
[...]

PR #222, opened 2026-09-04
fix(commons): the request log gets a way out (#221)

### LINKED ISSUE #221: commons: the request log has no way out — it is written by every request, read by nothing, and never deleted
Reported from a downstream fork that measured it on real hardware. The code is identical here, so the numbers transfer.

## Nothing removes a request log, ever

`log_request` gets a document per request, from middleware every service registers. The index declaration for it (`account/app/database/database_setup.py:84`) carries one entry and no `expireAfterSeconds`:

```python
index_key_list = [
    {"keys": {...}, "unique": False},
]
```

#182 fixed which fields that index is keyed on. It did not give the collection a way to shrink, and there is no other one — no TTL, no sweep, no retention job. `cart` has a TTL on its own collection, so the mechanism exists; `log_request` just never got one.

## And nothing reads it either

`RequestLogRepository` is defined and instantiated by nobody outside its own module. The collection is written by every request and queried by no code path in the tree.

That is a defensible design for an audit trail somebody reads out-of-band. It is not defensible together with unbounded growth: the thing accumulating has no consumer to justify the space.

## What it costs

Measured on an edge machine in the fork — a store PC, not a server:

| | |
|---|---|
| documents | **6,389,515** |
| size | **23.80 GiB** |
| share of the whole database | **97%** |

The volume itself slowed checkout down. On a store terminal there is nowhere for 23 GiB to go.

## What the fork did, and the order that matters

A TTL index on `created_at`, declared beside the retired unique index. Three things bit while getting there, and they are properties of the code rather than of that fork:

**The unique index has to be retired before startup provisioning runs.** `create_collection_with_indexes_async` raises as soon as one declared index cannot be built on an existing collection. While the unique index is still declared, a service whose request log already holds colliding rows fails to **start**, rather than merely losing records.

**The TTL has to be on a different key pattern from the retired index.** A unique index cannot be swapped for a non-unique one on the same keys: the end-state check compares key patterns without looking at uniqueness, so the replacement reads as "the stale index is still there" and provisioning raises.

**Declarations only reach a tenant created afterwards.** Provisioning runs at tenant creation, so an existing environment needs the index applied separately — and changing the TTL later needs the same, because the index ensure compares neither `expireAfterSeconds` nor any other option and silently keeps the old value.

## What would close it

A TTL on `created_at` with the retention period in settings, declared where the other request-log index is. The three constraints above are what a migration has to respect; they are not obvious from the code and cost a working day to rediscover.

If the intent is that the request log is *not* an audit trail and can simply be dropped, that would close it too — but then the write on every request is the thing to remove instead.

kugel-masa:
Closes #221.

`log_request` took a document per request, from middleware every service registers, and nothing ever removed one. Measured downstream on store hardware: **6,389,515 documents, 23.80 GiB, 97% of the database.** The volume itself slowed checkout down, and on a store PC there is nowhere for 23 GiB to go.

A TTL on `created_at`, retention in settings. Getting there meant fixing five things that were not true.

## 1. The date did not exist

`RequestLog` inherits `created_at` from `AbstractDocument`, but the only thing that ever set it is `AbstractRepository.create_async` — and the request-log buffer deliberately does not go through the repository: it batches across tenants and writes with `insert_many`, which is what keeps request logging off the latency path (#17).

So every request log was written with `created_at` null. A TTL on it would have installed cleanly and deleted **nothing**, and the next sign of trouble would have been the disk filling again. Stamped by a `default_factory` on the document rather than at the two call sites in the middleware, so a third one cannot be born without a date.

## 2. Half the volume was never provisioned

Every record is written twice — to the tenant database and to `{prefix}_commons` — but `create_request_log_collection` only ever ran for the tenant. The commons copy had no index beyond `_id_` and no way to shrink, while taking the same traffic. Now provisioned for both.

## 3. `tenant_id=None` did not mean what it means

That is how cart and terminal address the commons database. The other five interpolated it into the name and produced `db_<service>_None` — created, indexed, and read by nothing. Found by running the provisioning and looking at what MongoDB actually held rather than at what the call said. Fixed in all five and pinned by a test.

## 4. A changed retention was silently ignored

`createIndexes` will not modify options on an existing index: it answers `IndexOptionsConflict`, the ensure loop logs a warning, and the end-state check — which compares keys and uniqueness — then reports success over the value the index was **first** created with. An operator shortening retention because a disk is filling got no error and no effect. The declared value is now applied with `collMod`; verified live, four commons databases moved from 604800s to the declared 2592000s on the next provisioning pass.

Where a same-key index exists **without** any expiry, `collMod` cannot add one, so the plain index is dropped and rebuilt from the declaration — and if that fails the migration raises rather than reporting an expiry nothing performs.

## 5. The rows already there were out of reach

*(External review caught this one.)* A TTL skips a document whose indexed field is not a date, so the millions of pre-existing null rows — the entire reason for the issue — would have survived the fix. The date is recovered from `_id`: an ObjectId embeds the second it was generated, and these documents are inserted within seconds of the request. One server-side pipeline update, before the index, only over `{"created_at": None}`; a failure is logged and setup carries on.

Verified live: 104,267 dateless rows in `db_cart_commons` came back dated, none left null.

## Retention

Default 30 days. The collection has no reader in this tree — it is an audit trail somebody queries out of band — so retention is a disk decision rather than a correctness one. The downstream fork runs 7 days on store hardware; that difference is deliberate.

`0` means the index is not declared at all. It is **not** passed through: `expireAfterSeconds: 0` means "expire at the stored date" to MongoDB, which would delete each request log almost as soon as it was written. *(Also from the review.)*

## Tests

Unit 8/8 services (commons 695, account 47, terminal 373, master-data 406, cart 697, report 229, journal 156, stock 225), integration 7/7, e2e 8/8.

Verified against the running stack rather than only asserted: the TTL index exists on both the tenant and commons copies of every service, every document carries a BSON date, the retention change is applied in place, the backfill reaches the old rows, and no `_None` database is created.

## Note for an existing deployment

Provisioning runs at tenant creation, so the index and the backfill reach an existing tenant only when its setup is re-run.

The other half of the issue — that nothing reads this collection either — is left open deliberately. This makes the trail bounded; whether to keep writing it at all is a separate decision.

https://claude.ai/code/session_01354h8PdbZCuE1BQkS5eLxd


kugel-masa:
## Review round 2, and the tests that were missing

Asked whether the testing and review were sufficient, they were not, in two specific ways: the branch added in round 1 — dropping and rebuilding a same-key index that has no expiry — had **never run**, in a test or against a server, while being able to raise and stop a service starting; and the round-1 fixes themselves had not been reviewed.

**Ten integration tests** now run the real provisioning against MongoDB (`services/account/tests/integration/test_request_log_retention.py`). Each was checked to fail without the code it covers — the commons pass removed, the backfill call removed, the rebuild branch disabled, the constraint check disabled, the race tolerance disabled, the retention-off guard disabled — one failure each, and all ten pass again afterwards.

One of those tests was itself wrong the first time: it patched `drop_index` on a collection object, and motor hands back a new object for every `db[name]`, so it passed while exercising nothing. It uses a proxy now, and the assertion that the race was actually reached is what caught it.

**Round 2 found three more P1s**, all valid:

- **A constraint the declaration lacks was dropped silently.** Rebuilding from the declaration removes a `unique` or `partialFilterExpression` the live index carries — widening a constraint while adding an expiry. It now stops and names the option.
- **A concurrent provisioning failed startup.** Two services setting up the same collection both see the plain index; the loser's `drop_index` finds nothing to drop and raised through. The end state is re-read instead: the declared keys with an expiry means someone else did the work.
- **Retention off still walked the collection.** With `REQUEST_LOG_TTL_SECONDS=0` the backfill ran anyway, a full pass over millions of rows to prepare for an index that is never declared.

One correction to that review: the backfill runs at **tenant creation**, not at every startup as the comment said. The fix is right regardless.

Commons 0.1.94. Unit 8/8 services, integration 7/7, e2e 8/8, all after a full rebuild.

